PR #20, opened 2023-10-26
Fixed:remove gpt-4 from the default teacher_runtimes

### LINKED ISSUE #19: Agent creation fails without GPT4
https://github.com/HumanSignal/Adala/blob/71179879c6f754193c6950613a0bd0fcdded87f0/adala/agents/base.py#L44-L49

The `teacher_runtimes` will call `OpenAIRuntime(model='gpt-4')` all the time when we new a `Agent` instance like the example in  Quickstart: 

https://github.com/HumanSignal/Adala/blob/71179879c6f754193c6950613a0bd0fcdded87f0/README.md?plain=1#L127-L153

The problem is that when the user doesn't have access to GPT4, they can not create any `Agent` for now. The raised errors like:

```python

Requested model "gpt-4" is not available in your OpenAI account. Available models are:
...

Try to change the runtime settings for OpenAIRuntime, for example:

OpenAIRuntime(..., model="gpt-3.5-turbo")
...

---> 51     raise ValidationError(f'Requested model {self.gpt_model_name} is not available in your OpenAI account.')

TypeError: No constructor defined
```

shenxiangzhuang:
Fix #19

### DIFF
--- a/adala/agents/base.py
+++ b/adala/agents/base.py
@@ -44,7 +44,7 @@ class Agent(BaseModel, ABC):
     teacher_runtimes: Optional[Dict[str, Runtime]] = Field(
         default_factory=lambda: {
             'openai-gpt3': OpenAIRuntime(model='gpt-3.5-turbo'),
-            'openai-gpt4': OpenAIRuntime(model='gpt-4')
+            # 'openai-gpt4': OpenAIRuntime(model='gpt-4')
         }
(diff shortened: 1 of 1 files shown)


niklub:
@shenxiangzhuang thanks for your contribution!

niklub (approved):

PR #47, opened 2023-12-11
Add fast runtime for batch processing

niklub:


README.md
```diff
@@ -96,6 +96,39 @@ export OPENAI_API_KEY='your-openai-api-key'
 
 ## 🎬 Quickstart
 
+1. Define agent reasoning workflow in `workflow.yml`:
```
deppp, line 99: I'm not sure why this should be the main entry point into Adala and even be present in the readme. To me that looks like a possible addition worth speaking about in the docs.

adala/agents/base.py
```diff
@@ -311,3 +314,36 @@ def learn(
                     break
 
         print_text("Train is done!")
+
+
+def create_agent_from_dict(json_dict: Dict):
```
deppp, line 319: not sure why we need that abstraction?

adala/cli.py
```diff
@@ -0,0 +1,91 @@
...
+    elif args.file_type == "dataset":
+        run_dataset(
+            agent, args.input, args.dataset_config, args.dataset_split, args.output
+        )
+    else:
+        raise ValueError(f"Unknown file type {args.file_type}")
+    print("Done!")
+
+
+def parse_arguments(args):
```
deppp, line 59: I think CLI should be a separate addition, i.e. PR, and designed properly from the ground up so that we don't end up introducing breaking changes.

adala/utils/gsheet.py
```diff
@@ -0,0 +1,63 @@
+import pandas as pd
+import gspread
+from google.auth import default
+
+
+def get_client():
```
deppp, line 6: If my assumption that this would require a public sheet is correct I think we may want to add auth creds soon.

adala/runtimes/_guidance.py
```diff
@@ -123,7 +118,7 @@ def record_to_record(
         instructions_template: str,
         output_template: str,
         extra_fields: Optional[Dict[str, Any]] = None,
-        field_schema: Optional[Dict] = None,
+        options: Optional[Dict] = None,
```
deppp, line 121: I think field_schema had more clarity behind it. Do we have anything new in "options" other than field_schema?

README.md
niklub: It's a tentative to get a more human-readable form of creating agents. We can remove it from readme, if it doesn't bring too much value, however it might be more convenient and more robust rather than copypasting and changing the python code (based on my experience of creating long reasoning chains)

adala/agents/base.py
niklub: for deserialization

adala/cli.py
niklub: ok, we can hide it from this PR, anything in mind for the specific CLI design?

adala/utils/gsheet.py
niklub: Every gsheet access will require auth - this one uses default credentials (i.e. accessing from GCP, colab or locally with creds stored in environmental variables)

adala/runtimes/_guidance.py
niklub: `field_schema` still remains as user facing parameter in skills. However, I'm unsure about what should be the proper runtime parameter, how customizable it has to be, why would we need broader definition of generations schema or is it too redundant, and only string input->string output is well enough for most of the use cases?

PR #50, opened 2024-01-06
A bunch of improvements for the classification skill

chemeris:
Thank you for an excellent library. I ran it against our dataset of 10k social media posts to detect their sentiment and classify it into a set of topics. This set of patches is what it took for me to get it working - from minor fixes to improvements in the learning process.

I have a bunch of ideas on how to improve learning performance with more advanced learning strategies - I'd be happy to discuss this if there is interest in implementing advanced learning strategies.

### DIFF
--- a/adala/environments/base.py
+++ b/adala/environments/base.py
@@ -207,9 +207,9 @@ class StaticEnvironment(Environment):
                 [gt_pred_match.rename("match"), gt], axis=1
             )
             pred_feedback[pred_column] = match_concat.apply(
-                lambda row: "Prediction is correct."
+                lambda row: f"Prediction for {gt_column} is correct."
                 if row["match"]
-                else f'Prediction is incorrect. Correct answer: "{row[gt_column]}"'
+                else f'Prediction for {gt_column} is incorrect. Correct answer: "{row[gt_column]}"'
                 if not pd.isna(row["match"])
                 else np.nan,
                 axis=1,
--- a/adala/agents/base.py
+++ b/adala/agents/base.py
@@ -305,6 +305,12 @@ class Agent(BaseModel, ABC):
                 if skill.frozen:
                     continue
 
+                if accuracy[skill_output] >= accuracy_threshold:
+                    print_text(
+                        f'Output {skill_output} of skill "{skill_name}" is already accurate enough ({accuracy[skill_output]}), skipping ...'
+                    )
+                    continue
+
                 print_text(
                     f'Skill output to improve: "{skill_output}" (Skill="{skill_name}")\n'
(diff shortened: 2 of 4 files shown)


adala/environments/base.py
```diff
@@ -207,9 +207,9 @@ def get_feedback(
                 [gt_pred_match.rename("match"), gt], axis=1
             )
             pred_feedback[pred_column] = match_concat.apply(
-                lambda row: "Prediction is correct."
+                lambda row: f"Prediction for {gt_column} is correct."
```
niklub, line 210: Could you explain the reason for this addition? the initial idea was to use a single column at a time, so pointing out a specific column name might be not necessary - but I'm probably missing your idea

niklub:
Hi, @chemeris ! thanks for your great contribution!
> I'd be happy to discuss this if there is interest in implementing advanced learning strategies.

Absolutely, it would be very helpful to have different learning strategies and reasoning path, and give user the options. Feel free to open the github issue where we can discuss the solutions and reference it in PR hereafter

adala/environments/base.py
chemeris: Classification (and Transform) skills support multiple outputs and I used this to classify each social media post into multiple categories (each one as a True/False field).

You're correct that on each step we evaluate only one output. But the base prompt doesn't mention which of the outputs we evaluate at the step. This patch is the easiest way I found to make sure the model understands that the feedback is related to a specific output.

If there is only one output, we can simply say "Prediction is correct." as it used to be.

Here is how the template from TransformSkill.improve(). Note that it doesn't mention anything about the output name:
```Python
"""
## Current prompt
{self.instructions}

## Examples
{examples}

Summarize your analysis about incorrect predictions and suggest changes to the prompt."""
```

niklub (approved):

adala/environments/base.py
niklub: I got your idea about referencing specific columns to help LLM make the correct assessment. However, the column name defined there doesn't contain any signal for LLM, example from the tests:
```
Prediction for gt_0 is incorrect. Correct answer:     0 0 0   1 1 1   1 5 1   
  "1 1 1" 
```
"gt_0" keyword is not presented in input prompt which consists of the string "Input: ... Output: ...". In this case, I'd better create a string like
`"Prediction for the field "Output" is incorrect"`
assuming there can be multiple outputs.
Let me know if it makes sense.
Happy to merge your PR as soon as we have all tests passed. Thank you.

niklub (changes requested):

adala/environments/base.py
chemeris: Makes sense.
I just need to figure out how to get the field name 🤔

PR #60, opened 2024-02-07
initial version of API server

deppp:


matt-bernstein:
@deppp should this branch be closed in favor of feat/async-env-kafka, which also implements an API server?

PR #78, opened 2024-04-01
fix: DIA-986: Upgrade OpenAI client version & pytest coverage

niklub:
- Openai client migrated to v1 version

- Tests and server mocks were fixed to address v1 compatibility

- Some tests were removed since they don’t add additional logic → targeting to increase coverage as soon as reports will be ready (2 more tests are coming: AsyncOpenAIChatRuntime, server API)


robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/4e9673adf2a3b7492671dee82abec7fb6ccb6cc8 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/e106b718eb505256dee848ae50733dd3b7f814d9 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/4ba535e75095a3a89fe8684a9a69460346ab63a5 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/b732534d30707b30b3fdf51c0185bc4ba6a6bcd4 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/0be413ab6e5883bb5d565e905a938ccad83948fd is created.

adala/runtimes/_openai.py
```diff
@@ -235,15 +185,10 @@ def execute(self, messages: List):
...
-            )
-            completion_text = completion.choices[0].message.content
-        else:
-            # deprecated
-            completion = chat_completion_call(self.openai_model, messages)
-            completion_text = completion.choices[0]["message"]["content"]
+        completion = self._client.chat.completions.create(
+            model=self.openai_model, messages=messages
+        )
+        completion_text = completion.choices[0].message.content
```
matt-bernstein, line 191: TY, this is much more readable now that it's not supporting both old and new openai formats

tests/utils.py
```diff
@@ -1,24 +1,42 @@
...
+
+
+@dataclass
+class OpenaiChatCompletionMessageMock(object):
+    content: str
+
+
+@dataclass
+class OpenaiChatCompletionChoiceMock(object):
+    message: OpenaiChatCompletionMessageMock
```
matt-bernstein, line 23: Do you have a general strategy for how to mock openai responses in the future? If we only care about chat completions, this seems ok to start with, but there must be libraries that handle this for more than one endpoint in a cleaner way, eg you've mentioned https://github.com/polly3d/mockai. Also compare to what we're doing in pytest for LSE: https://github.com/HumanSignal/label-studio-enterprise/blob/develop/label_studio_enterprise/lse_tests/utils.py#L221

adala/runtimes/_guidance.py
matt-bernstein, line 1: Definitely happy with reducing the code surface area, but there other runtimes planned to support constrained generation? https://github.com/jxnl/instructor seems like a pretty solid low-overhead one to set up for example

matt-bernstein (approved):

adala/runtimes/_guidance.py
niklub: It's premature to extend it with different constrained generation frameworks until the agent workflow is well-established and tested. But overall, it would be great to follow a minimalistic high-level API to define constrained generation schema, similar to the example

tests/utils.py
niklub: It makes sense to have a single testing tool to mock LLM clients. We can use them to `label-studio-sdk` for example, to reuse across different apps

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/b818d466aaf0b99104a461b68710337386df4659 is created.

PR #98, opened 2024-04-23
Running agent.arun(input=df) with environment=None is totally valid.

chemeris:
Otherwise this code requires us to create an dummy async environment which is not used in the code and not even possible with the current code base.

### DIFF
--- a/adala/agents/base.py
+++ b/adala/agents/base.py
@@ -259,13 +259,13 @@ class Agent(BaseModel, ABC):
         else:
             print(f"Using runtime {type(runtime)}")
 
-        if not isinstance(self.environment, AsyncEnvironment):
-            raise ValueError(
-                "When using asynchronous run with `agent.arun()`, the environment must be an AsyncEnvironment."
-            )
         if input is None:
             if self.environment is None:
                 raise ValueError("input is None and no environment is set.")
+            if not isinstance(self.environment, AsyncEnvironment):
+                raise ValueError(
+                    "When using asynchronous run with `agent.arun()` and noe input, the environment must be an AsyncEnvironment."
+                )
             # run on the environment until it is exhausted
             while True:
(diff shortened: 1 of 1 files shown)


niklub (approved):
Thanks for the fix. Although I totally agree with your proposal, it is still interesting how do you specifically use async call without async environment? It would be great if you could provide an example code so we would incorporate it in our future tutorials

chemeris:
Thank you, and I look forward to having this merged.

I can't provide an example code, unfortunately, but the code we're writing is based on the `classification_skill.ipynb` example. We simply changed the synchronous `agent.run()` to asynchronous `await agent.arun()`. 

Two reasons for this:
1. We're running in async FastAPI handler anyway, so why not?
2. The async agent supports setting `temperature=0`, while sync one doesn't. And given the default temperature is AFAIK set to 1, it reduces reliability of the classification. And in general it looked like sync code is pretty much abandoned in favor for the new async one.

That said, I always thought that an `environment` is only needed during "teaching". Are you saying there are other use cases for it during normal production operation? The documentation is not very detailed on this part.

chemeris:
Hi, is there anything else required to merge this PR?

chemeris:
@niklub Please, could you merge this PR?
I see that the change is approved but it's still says "This workflow requires approval from a maintainer"

chemeris:
Any news?

chemeris:
Hi, ping about this PR - it's really simple

chemeris:
@niklub @matt-bernstein Any chance this can be merged?
It's a really simple one and @niklub has approved, but someone with write access to the repo has to click the button to merge.

matt-bernstein:
Thanks for the nudge here. Appreciate not letting this one fall through the cracks. :slightly_smiling_face:

PR #105, opened 2024-05-07
Fix for handing errored out openai response to not include them in response packet sent to LSE

forum-hs:


robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/9c3f31d796c957eb8e006269f8a88216132a509b is created.

adala/runtimes/_openai.py
```diff
@@ -382,11 +387,13 @@ async def batch_to_batch(
                     if name in options:
                         completion_text = match_options(completion_text, options[name])
                     outputs.append({name: completion_text})
+                    index.append(idx)
```
pakelley, line 390: can we just include the index in the `outputs` so we don't have to manage a separate iterable? So just do something like `outputs.append({name: completion_text, "index": idx})`

adala/runtimes/_openai.py
```diff
@@ -363,10 +363,15 @@ async def batch_to_batch(
                 )
 
                 # parse responses, optionally match it with options
-                for prompt, response in zip(prompts, responses):
+                index = []
+                for prompt, response, idx in zip(prompts, responses, batch.index):
                     # check for errors - if any, append to outputs and continue
                     if response.get("error"):
-                        outputs.append(response)
+                        # FIXME if we collect failed and succeeded outputs in the same list -> df, we end up with an awkward schema like this:
```
pakelley, line 370: Also, can we explicitly state here that we're intentionally not appending to `outputs` here? I know it's clear in context, but I could still see someone thinking this was a mistake if we don't explicitly explain

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/248ab9ae131a663c272469913d1d58867ae8c180 is created.

hakan458 (approved):

PR #107, opened 2024-05-08
Create and clean up kafka topics manually to reduce memory burden for Adala server

matt-bernstein:


robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/7fafaf80164b9a0a8dbd6a8ae1a7125c4328ee5f is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/129c5bc7a8729dd74b64383b11390c224a201081 is created.

adala/environments/kafka.py
```diff
@@ -61,12 +61,12 @@ async def message_receiver(self, consumer: AIOKafkaConsumer, timeout: int = 3):
                 try:
                     # Wait for the next message with a timeout
                     msg = await asyncio.wait_for(consumer.getone(), timeout=timeout)
-                    # print_text(f"Received message: {msg.value}")
+                    print_text(f"Received message: {msg.value}")
```
hakan458, line 64: If we're going to keep these around lets make it more clear where the log is coming from - because we have consumers in so many different parts of the code its not clear when looking at logs

server/tasks/process_file.py
```diff
@@ -116,6 +116,9 @@ def process_file_streaming(self, agent: Agent, parent_job_id: str):
     # Set input and output topics using parent job ID
     agent.environment.kafka_input_topic = get_input_topic(parent_job_id)
     agent.environment.kafka_output_topic = get_output_topic(parent_job_id)
+    import time
+
+    time.sleep(5)
```
hakan458, line 121: why??

server/utils.py
```diff
@@ -18,8 +21,68 @@ class Settings(BaseSettings):
 
 
 def get_input_topic(job_id: str):
-    return f"adala-input-{job_id}"
+    topic_name = f"adala-input-{job_id}"
+
+    # same logic as get_output_topic
+
+    settings = Settings()
```
hakan458, line 28: Can we separate out the logic for topic creation into its own function to not have a duplicate? get_input / output can just call that function

adala/environments/kafka.py
matt-bernstein: oops, this was for testing, will revert

server/tasks/process_file.py
matt-bernstein: also for testing, will revert

server/utils.py
matt-bernstein: yep

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/1c746fad389ccce43816e0be29288e4fb72df292 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/14380a999c9abcee342c2b9f640c7d4936626af1 is created.

adala/environments/kafka.py
```diff
@@ -61,12 +61,12 @@ async def message_receiver(self, consumer: AIOKafkaConsumer, timeout: int = 3):
...
                     yield msg.value
                 except asyncio.TimeoutError:
-                    # print_text(
-                    # f"No message received within the timeout {timeout} seconds"
-                    # )
+                    print_text(
+                        f"No message received within the timeout {timeout} seconds"
+                    )
```
forum-hs, line 69: same as above

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/3f06167275b421d788aeacb247e9f7c6acc11628 is created.

server/tasks/process_file.py
```diff
@@ -117,9 +117,16 @@ def process_file_streaming(self, agent: Agent, parent_job_id: str):
     agent.environment.kafka_input_topic = get_input_topic(parent_job_id)
     agent.environment.kafka_output_topic = get_output_topic(parent_job_id)
 
+    # import time
+    # time.sleep(5)
```
forum-hs, line 121: remove this

server/tasks/process_file.py
```diff
@@ -117,9 +117,16 @@ def process_file_streaming(self, agent: Agent, parent_job_id: str):
...
+
     # Run the agent
     asyncio.run(agent.arun())
 
+    # manually cleanup these topics
+    delete_topic(agent.environment.kafka_input_topic)
+    delete_topic(agent.environment.kafka_output_topic)
```
hakan458, line 128: Should we do this in the parent task instead of this input task?

matt-bernstein: I want to keep it as close to where they are used as possible, we already need to get the topic name here so no sense doing it in 2 places. It could also be different for different subtasks called by the same parent task in the future (for ex, when we switch to a reusable topic pool it'll be "unlocking" the topic instead of deleting it)

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/c227c64baa6666dac855df947bba18a4530459ca is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/e2094004ad3e9f662f712802cce44fac17fe8b02 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/2193a9038b487f0868d6f067720579bbe45f53d5 is created.

server/tasks/process_file.py
hakan458: I think there is a risk here - input job is done, we delete the topic, but output job is still trying to run/pull things from output topic queue, so I would not delete it here

matt-bernstein: oh you're right actually, I'll change it

server/utils.py
```diff
@@ -18,8 +21,50 @@ class Settings(BaseSettings):
 
 
 def get_input_topic(job_id: str):
-    return f"adala-input-{job_id}"
+    topic_name = f"adala-input-{job_id}"
+
+    ensure_topic(topic_name)
```
pakelley, line 26: Imo it might be better to only return the input topic name here, and leave it to the user to ensure the topic exists. Seems more "separation of concerns" to me.
I'm pretty on the fence though, since I can't think of a use case where you'd want the name of a non-existent topic (or not-yet-existent/no-longer-existent topic), so I'll leave it up to your judgment.

server/tasks/process_file.py
hakan458: Not changed yet right?

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/b1c6f45809a2360c0966d0e3536ddea55e4d8648 is created.

server/tasks/process_file.py
matt-bernstein: done

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/c23b8f3b35ebef3b631628535cd8c16125bf8ddc is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/9fcff5cdbcfcee52e1705435fe4f8703aa616480 is created.

pakelley (approved):

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/7c3beca199ab1c539f3bacd783d81f32cf70b2b0 is created.

PR #120, opened 2024-05-22
feat: DIA-1082: task-level error handling

matt-bernstein:
expose errors coming from OpenAI instead of ignoring them in `AsyncOpenAIRuntime`, with several knock on effects - see commit messages b61fc01d069cee26dc2cc42cb30635cfb6237b30 and later.

(Earlier commits are from another branch and already in master)

- use exponential backoff retries
- use openai async client instead of requests
- reuse client across batches
- always return error fields even when empty
- expose consumer timeouts in Settings object
- improved kafka handling in KafkaEnvironment to match best practices in server

A successful result task now always looks like

```
        {
<fields from input record, eg text, task_id>
            "output": output,
            "_adala_error": False,
            "_adala_message": None,
            "_adala_details": None,
        }
```
and an unsuccessful result task looks like
```

        {
<fields from input record, eg text, task_id>
            "output": None,
            "_adala_error": True,
            "_adala_message": "Unknown error",
            "_adala_details": str(e),
        }

```

Error message and details should be special-cased as we encounter them and decide to handle them differently.


robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/1ee8eeabd5d2995b3e6e375d8d20b152d5b68f07 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/f33eeaea69a86346368e3b584506f3454f72ff89 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/4149336f9b193a611a421eaee1d524f7cb0295ed is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/9e25b525ec04f06cdd548031ad1cb40a0df6fd95 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/4718f60fe24be2f4cbd96755715dd6b3f6e6f8f1 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/8e5df9f94b3afeecc62ec3741392e015ae2521c7 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/e7c98486cf772457a7faf6250d15fed6796a21d8 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/20a304396c5fcaec24d2a5e23631b182d3033c47 is created.

adala/environments/kafka.py
```diff
@@ -26,19 +26,43 @@ class AsyncKafkaEnvironment(AsyncEnvironment):
...
         kafka_input_topic (str): The Kafka input topic.
         kafka_output_topic (str): The Kafka output topic.
+        timeout_ms (int): The timeout for the Kafka consumer.
     """
 
-    kafka_bootstrap_servers: Union[str, List[str]]
-    kafka_input_topic: str
-    kafka_output_topic: str
+    # these are mandatory, but should be set by server
+    kafka_bootstrap_servers: Optional[Union[str, List[str]]] = None
```
pakelley, line 33: Could we add these to an __init__ function or something? I think this workflow is a little confusing.

adala/environments/kafka.py
```diff
@@ -26,19 +26,43 @@ class AsyncKafkaEnvironment(AsyncEnvironment):
...
+    timeout_ms: Optional[int] = None
+
+    # these are set in initialize()
+    consumer: Optional[AIOKafkaConsumer] = None
+    producer: Optional[AIOKafkaProducer] = None
 
     async def initialize(self):
-        # claim kafka topic from shared pool here?
-        pass
+        assert self.kafka_bootstrap_servers is not None, "missing initialization"
```
pakelley, line 43: If we do have to keep with this workflow, we should at least state which param is missing in these asserts

adala/runtimes/_openai.py
```diff
@@ -64,86 +54,53 @@ async def async_create_completion(
...
-    except aiohttp.ClientError as e:
-        # Handle other aiohttp specific errors
-        return {
-            "error": True,
-            "message": "Client error",
-            "details": str(e),
-        }
-    except asyncio.TimeoutError as e:
-        # Handle timeout errors
+        completion = await client.chat.completions.create(
```
pakelley, line 57: just so happy to see this change, had to comment lol

adala/runtimes/_openai.py
```diff
@@ -64,86 +54,53 @@ async def async_create_completion(
...
+            max_tokens=max_tokens,
+            temperature=temperature,
+        )
+        completion_text = completion.choices[0].message.content
         return {
-            "error": True,
-            "message": "Request timed out",
-            "details": str(e),
+            "text": completion_text,
+            "__adala_error__": False,
```
pakelley, line 66: nit: I think it's more common to denote private "attributes" in python with a single "__" before the name (as opposed to private/reserved methods with the "__" on both sides)

adala/runtimes/_openai.py
```diff
@@ -64,86 +54,53 @@ async def async_create_completion(
...
-                    session=session,
-                    model=openai_model,
-                    max_tokens=max_tokens,
-                    temperature=temperature,
-                    instruction_first=instruction_first,
-                    default_timeout=timeout,
-                    openai_api_key=openai_api_key,
-                )
+    tasks = []
+    for prompt in prompts:
```
pakelley, line 89: this loop seems simple enough that it may as well be a list comprehension. If you prefer the loop (maybe because of the large function signature), all good, but figured I'd suggest it.

adala/environments/kafka.py
matt-bernstein: The reason they're not in an init is that the Agent and Environment get instantiated inside the app's `/submit-streaming` endpoint handler due to the function signature there, before we have access to the Settings() object which contains these fields. Could modify that endpoint handler's pydantic Request object to special-case these fields, but that seems like too much magic to reasonably expect the server's user to know about.

adala/runtimes/_openai.py
matt-bernstein: done, good call

adala/runtimes/_openai.py
matt-bernstein: done, good call

adala/environments/kafka.py
matt-bernstein: done

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/8fe78f7e2906ddb2358bddfc4186522ed82030c9 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/83e86104a676dac4f47052c8035426dd11298707 is created.

adala/environments/kafka.py
pakelley: ugh, gotcha, feels like these abstractions are a bit too cute, but as far as this change goes it makes sense to keep it as-is 👍

server/handlers/result_handlers.py
```diff
@@ -95,6 +95,26 @@ def ready(self):
...
+        known_keys = [
+            # 'text',  # should always be omitted
+            "task_id",
+            "output",
+            # TODO in DIA-1122
+            # '_adala_error',
+            # '_adala_message',
+            # '_adala_details',
+        ]
+        result_batch = [
```
pakelley, line 113: seems like a good place to have pydantic types for validation, maybe?

pakelley (approved):
Would like a pydantic type for validating our returned results, but not a blocker

adala/environments/kafka.py
matt-bernstein: yep, for sure we should rethink and de-abstract some of this for the server, if not adala in general

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/37ebe4908d8dcb3938d45f9bf0fcb40670354fcf is created.

server/handlers/result_handlers.py
matt-bernstein: done

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/8db422cfec6e3f89c2c7c3ef53ffb8b288dc7f58 is created.

pakelley (approved):

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/7742415a2c242fa9d0f5470f53e7e536beecd9b1 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/75183988a6ce733a6d54ab729208ea2003c792f7 is created.

chemeris:
Hi @matt-bernstein 
This change breaks generation for non-categorical output types (e.g. email generation). Specifically, this line is the issue:
https://github.com/HumanSignal/Adala/commit/4679bfbc287036f14950ccd78ebcb81a0c7ed980#diff-87a8807ecc3037662d3036e959db10ec4479f616ccf0fcd62ee2aaa9a6cdcb5cR349

`name in options` is only `True` for categorical variables, but is never `True` for freeform text generation, so `response[name] = completion_text` is not executed and we get a response with no variable in it. `response[name] = completion_text` must be outside of the `if` statement and always executed.

I had to fix this for our system to continue operating: https://github.com/chemeris/Adala/commit/5448e37567801ac5a565774024e5a85b5818fd13

matt-bernstein:
> Hi @matt-bernstein This change breaks generation for non-categorical output types (e.g. email generation). Specifically, this line is the issue: [4679bfb#diff-87a8807ecc3037662d3036e959db10ec4479f616ccf0fcd62ee2aaa9a6cdcb5cR349](https://github.com/HumanSignal/Adala/commit/4679bfbc287036f14950ccd78ebcb81a0c7ed980#diff-87a8807ecc3037662d3036e959db10ec4479f616ccf0fcd62ee2aaa9a6cdcb5cR349)
> 
> `name in options` is only `True` for categorical variables, but is never `True` for freeform text generation, so `response[name] = completion_text` is not executed and we get a response with no variable in it. `response[name] = completion_text` must be outside of the `if` statement and always executed.
> 
> I had to fix this for our system to continue operating: [chemeris@5448e37](https://github.com/chemeris/Adala/commit/5448e37567801ac5a565774024e5a85b5818fd13)

Thanks for the quick feedback @chemeris ! Merged in your change.

PR #131, opened 2024-06-11
fix: DIA-1168: Delete kafka topics on job failure / cancel

hakan458:
We observed an issue where the list of kafka topics was growing on develop Adala instance, even though we implemented logic to delete them after completing a job. However, we did not handle the case of a job failing, or being canceled. I have added a handler to be run when the parent_streaming_task fails, so that we clean up topics, and also cleaning up topics when we cancel a job.

Also added logic to attempt to stop input/output jobs when the parent task fails.

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -25,7 +25,7 @@ from tasks.process_file import (
     process_streaming_output,
     streaming_parent_task,
 )
-from utils import get_input_topic_name, Settings
+from utils import get_input_topic_name, get_output_topic_name, Settings
 from server.handlers.result_handlers import ResultHandler
 
 
@@ -284,11 +284,13 @@ def get_status(job_id):
         "RETRY": Status.INPROGRESS,
     }
     job = streaming_parent_task.AsyncResult(job_id)
-    logger.info(f"\n\nParent task meta : {job.info}\n\n")
+    logger.info(f"Parent task meta : {job.info}")
 
     # If parent task meta does not contain input/output job IDs - return FAILED
     if (
         job.info is None
+        or type(job.info)
+        != dict  # In some failure cases e.g. an exception is thrown, job.info will be a string, causing the next line to crash
         or "input_job_id" not in job.info
         or "output_job_id" not in job.info
     ):
@@ -341,6 +343,13 @@ def cancel_job(job_id):
             f"output job id: {input_job_id} unavailable to cancel for parent job: {job_id}"
         )
     job.revoke()
+
+    # Delete Kafka topics
+    input_topic_name = get_input_topic_name(job_id)
+    output_topic_name = get_output_topic_name(job_id)
+    delete_topic(input_topic_name)
+    delete_topic(output_topic_name)
+
     return Response[JobStatusResponse](data=JobStatusResponse(status=Status.CANCELED))
 
 
--- a/server/tasks/process_file.py
+++ b/server/tasks/process_file.py
@@ -29,6 +29,39 @@ app = Celery(
 )
 
 
+def parent_job_error_handler(self, exc, task_id, args, kwargs, einfo):
+    """
+    This function will be called if streaming_parent_task fails, to ensure that we cleanup any left over
+    Kafka topics.
+
+    Also attempts to stop input/output jobs if their IDs are available
+    """
+    parent_job_id = task_id
+    input_topic_name = get_input_topic_name(parent_job_id)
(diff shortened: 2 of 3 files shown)


robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/3144ea8053f8c9a81c06084acccd9bd9987a7da8 is created.

server/app.py
```diff
@@ -284,11 +284,13 @@ def get_status(job_id):
...
     }
     job = streaming_parent_task.AsyncResult(job_id)
-    logger.info(f"\n\nParent task meta : {job.info}\n\n")
+    logger.info(f"Parent task meta : {job.info}")
 
     # If parent task meta does not contain input/output job IDs - return FAILED
     if (
         job.info is None
+        or type(job.info)
+        != dict  # In some failure cases e.g. an exception is thrown, job.info will be a string, causing the next line to crash
```
hakan458, line 293: `black` linter made this a new line, looks weird but guess it works

matt-bernstein: make sure you're running `poetry run black` instead of just `black` so you don't accidentally use a local version with different linting rules. Happened to me

server/utils.py
```diff
@@ -65,4 +65,12 @@ def delete_topic(topic_name: str):
...
-    admin_client.delete_topics(topics=[topic_name])
+    fs = admin_client.delete_topics(topics=[topic_name])
+    print(fs)
+    # print(fs.__dict__)
+    # for topic, f in fs.items():
+    #     try:
+    #         f.result()
+    #         print("Topic {} deleted".format(topic))
+    #     except Exception as e:
+    #         print("Failed to delete topic {} {}".format(topic_name, e))
```
matt-bernstein, line 76: remove this or make it a debug log?

matt-bernstein (approved):

server/app.py
hakan458: Ah I def did that lol good call

server/utils.py
hakan458: Whoops yeah will delete, was debugging

server/app.py
hakan458: No changes with running that way fyi

server/utils.py
hakan458: Done

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/7cbe0c3eddada655513282936dc6fea28afb2136 is created.

forum-hs (approved):

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/663b690ce0818ba4dfb3de19b863067dcca29290 is created.

server/tasks/process_file.py
```diff
@@ -29,6 +29,39 @@
...
+    output_topic_name = get_output_topic_name(parent_job_id)
+    delete_topic(input_topic_name)
+    delete_topic(output_topic_name)
+
+    parent_job = streaming_parent_task.AsyncResult(parent_job_id)
+    if (
+        parent_job.info is None
+        or type(parent_job.info)
+        != dict  # In some failure cases e.g. an exception is thrown, job.info will be a string, causing the next line to crash
+        or "input_job_id" not in parent_job.info
```
pakelley, line 50: Can there be a case where we only have the input job, or only have the output job? If so, I figure we should still try to delete the one that we have

server/tasks/process_file.py
```diff
@@ -29,6 +29,39 @@
...
+        or "output_job_id" not in parent_job.info
+    ):
+        logger.warning(
```
pakelley, line 53: Can there be a case where we only have the input (or only have the output) job id? If so, I figure we should still try to stop the one we have access to.

hakan458: Discussed in Slack, we add both job's IDs to the parent task's metadata at the same time

pakelley: Discussed this w/ Hakan, and it'd be nice to handle the case when we (for example) start the input job successfully, but don't start the output job successfully. Ideally, we should cancel both the input and parent job in this case (to avoid using up the user's openai quota on predictions that will never make it back to LSE).
Since this is more in line with what Matt is working on, we can handle it as part of his work though.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/cd55f815a181208f672d8246a88d637ed64bf32d is created.

PR #135, opened 2024-06-14
Add server-side timeout

matt-bernstein:
Add server-side timeout to all Celery tasks. We already have an LSE timeout of 30 minutes, but this will catch individual Celery tasks that are orphaned (don't have a parent task available to cancel them) due to <3 workers being available when the job is started. 

Since the timeout applies to in progress jobs (that are waiting on sending/receiving kafka messages), setting a very loose default timeout of 24 hours.

### DIFF
--- a/server/tasks/process_file.py
+++ b/server/tasks/process_file.py
@@ -28,6 +28,7 @@ app = Celery(
     "worker", broker=REDIS_URL, backend=REDIS_URL, accept_content=["json", "pickle"]
 )
 
+settings = Settings()
 
 def parent_job_error_handler(self, exc, task_id, args, kwargs, einfo):
     """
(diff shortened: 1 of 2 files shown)


matt-bernstein:
A change to the way the `Settings` object works here (due to needing to pass settings to the @app decorators) is to have one singleton per file, instead of being reinstantiated in every function that needs it. This means that if a setting (env var) is changed during runtime, it will not take effect until the server is restarted. I don't think we were relying on this, but worth mentioning in case it doesn't seem worth it.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/a9f108b8252dde69c8ccd3c46bd2c4a436dd4940 is created.

pakelley (approved):
lgtm!

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/e73a2b503c4f327fafac0ae86ca15e50971dda8d is created.

PR #156, opened 2024-07-10
feat: DIA-1255: Use litellm for openai inference

pakelley:


server/handlers/result_handlers.py
```diff
@@ -75,7 +75,7 @@ class LSEBatchItem(BaseModel):
     )
 
     task_id: int
-    output: str
+    output: Optional[str]
```
pakelley, line 78: Notably, this is `None` when there's an error, and it ends up surfacing as a pretty cryptic pickling error (at least at the final stacktrace, there are logs beforehand that show the problem)
Making this Optional makes this a bit easier to trace (and we'll need it anyway once we improve our error handling)

adala/runtimes/_openai.py
```diff
@@ -42,16 +43,24 @@ async def async_create_completion(
...
+    messages = [{'role': 'user', 'content': user_prompt}]
     if system_prompt:
         if instruction_first:
             messages.insert(0, {"role": "system", "content": system_prompt})
         else:
             messages[0]["content"] += system_prompt
 
     try:
-        completion = await client.chat.completions.create(
+        # FIXME: I think we're better off doing this globally, since this
```
pakelley, line 57: I have some thoughts on how we could do this, and have tested an implementation, but def want to discuss more before pulling the trigger here.

adala/runtimes/_openai.py
```diff
@@ -42,16 +43,24 @@ async def async_create_completion(
     Returns:
         Dict[str, Any]: OpenAI response or error message.
     """
-    openai_api_key = openai_api_key or os.getenv("OPENAI_API_KEY")
-    messages = [{"role": "user", "content": user_prompt}]
+    openai_api_key = (
+        openai_api_key or client.api_key or os.environ['OPENAI_API_KEY']
+    )
+    messages = [{'role': 'user', 'content': user_prompt}]
```
pakelley, line 49: I know we've talked a bit about this, but is anyone opposed to me just switching adala over to blue?

forum-hs: +1

adala/runtimes/_litellm.py
```diff
@@ -0,0 +1,537 @@
...
+logger = logging.getLogger(__name__)
+
+
+@retry(wait=wait_random_exponential(min=1, max=60), stop=stop_after_attempt(6))
+async def async_create_completion(
+    model: str,
+    user_prompt: str,
+    client: httpx.Client,
+    system_prompt: Optional[str] = None,
+    api_key: Optional[str] = None,
```
pakelley, line 27: not actually used here, still have some cleaning to do...

matt-bernstein: How does litellm handle retries? does it have a better way to respect rate limits built in?

adala/runtimes/_litellm.py
```diff
@@ -0,0 +1,537 @@
...
+        arbitrary_types_allowed=True
+    )  # for @computed_field
+
+    model: str
+    api_key: Optional[str] = None
+    base_url: Optional[str] = None
+    max_tokens: Optional[int] = 1000
+    temperature: Optional[float] = 0.0
+    splitter: Optional[str] = None
+    # concurrent_clients: Optional[int] = 10
```
pakelley, line 266: left the concurrent clients stuff around in comments until I get around to testing more thoroughly whether we need it or not.
If anyone knows a reason we definitely still need it, or has a great way to test whether we need it, lmk.
Otherwise I'll be doing some kinda broad testing trying to elicit an issue.

adala/runtimes/_litellm.py
```diff
@@ -0,0 +1,537 @@
...
+                                "image": {
+                                    "type": "string",
+                                    "format": "uri"
+                                }
+                            }
+                            ```
+            instructions_first: If True, instructions will be sent before input.
+        """
+
+        # if not check_if_new_openai_version():
```
pakelley, line 462: left this just so I remember to include this in the openai vision runtime, fwiw

adala/runtimes/_openai.py
```diff
@@ -230,144 +141,12 @@ class AsyncOpenAIChatRuntime(AsyncRuntime):
...
-        record: Dict[str, str],
-        input_template: str,
-        instructions_template: str,
-        output_template: str,
-        extra_fields: Optional[Dict[str, Any]] = None,
-        field_schema: Optional[Dict] = None,
-        instructions_first: bool = True,
-    ) -> Dict[str, str]:
-        raise NotImplementedError("record_to_record is not implemented")
+    model: str
```
pakelley, line 148: forgot to go back and add an openai_model alias, TODO (same for api_key)

adala/runtimes/_litellm.py
matt-bernstein: If this works, it could resolve our ticket about rate limiting... https://docs.litellm.ai/docs/routing#advanced---routing-strategies-%EF%B8%8F

adala/runtimes/_litellm.py
hakan458: Without knowing litellm specifics - I figure concurrent clients will help us get results faster right? As long as rate limits are respected I would think 10 clients would be faster than 1, especially for larger projects. Depends on the rate that we are able to send tasks from LSE to Adala of course

adala/runtimes/_litellm.py
```diff
@@ -0,0 +1,537 @@
...
+    api_key: Optional[str] = None
+    base_url: Optional[str] = None
+    max_tokens: Optional[int] = 1000
+    temperature: Optional[float] = 0.0
+    splitter: Optional[str] = None
+    # concurrent_clients: Optional[int] = 10
+    timeout: Optional[int] = 10
+
+    # @computed_field
+    # def _client(self) -> AsyncOpenAI:
```
hakan458, line 270: can we get rid of this?

adala/runtimes/_litellm.py
pakelley: saw that, was hoping the same but have not had the chance to test. Probably going to need to test in a separate ticket at this point, but seems promising

adala/runtimes/_litellm.py
pakelley: sure, but do we ever actually create additional clients? Seems we only use this to pass is to httpx, but I could be missing some adala magic here

adala/runtimes/_litellm.py
pakelley: yep ofc, just left it as a reference for now until I commit to getting rid of concurrent clients

adala/runtimes/_litellm.py
matt-bernstein: this is legacy from before httpx was used to handle clients and each client was created manually. I don't think there's a benefit to controlling this at the user facing level.

tests/test_server.py
```diff
@@ -347,17 +347,14 @@ async def test_streaming_openai_only(openai_mock):
...
     )
 
-    from adala.runtimes._openai import async_create_completion
-    from openai import AsyncOpenAI
+    from adala.runtimes._litellm import async_create_completion
 
-    openai = AsyncOpenAI(
-        api_key=OPENAI_API_KEY,
-    )
+    http_client = httpx.Client()
```
matt-bernstein, line 352: unused? you can use the `client` and `async_client` fixtures

adala/runtimes/_openai.py
```diff
@@ -1,224 +1,46 @@
...
         base_url: Can point to any implementation of the OpenAI API. Defaults to OpenAI's.
         max_tokens: Maximum number of tokens to generate. Defaults to 1000.
     """
 
-    model_config = ConfigDict(arbitrary_types_allowed=True)  # for @computed_field
-
-    openai_model: str = Field(alias="model")
-    openai_api_key: Optional[str] = Field(
-        default=os.getenv("OPENAI_API_KEY"), alias="api_key"
+    model: str
```
pakelley, line 29: ah, yeah this is somewhat questionable. We're renaming this field here (bc I want it to override the litellm one, and it doesn't make sense to name it "openai_model" in litellm...)
We could add an alias (not a pydantic one, just a separate attr), but... idk it feels like an annoying thing to support, and seems unlikely anyone's using it.
Wdyt?

pakelley: Same goes for api_key, btw

adala/runtimes/_openai.py
```diff
@@ -1,224 +1,46 @@
...
+    # openai_model: str = Field(alias='model')
+    # TODO does it make any sense for this to be optional?
```
pakelley, line 31: Also, this question: seems like api_key should be required for an openai model, right? Whether we get it from the env var or it's passed in, we should be sure we have one before trying to do anything with openai.
Does that seem right?

adala/runtimes/_litellm.py
```diff
@@ -0,0 +1,488 @@
...
+    def init_runtime(self) -> 'Runtime':
+        # check model availability
+        try:
+            if self.api_key:
+                litellm.check_valid_key(model=self.model, api_key=self.api_key)
+        except NotFoundError:
+            raise ValueError(
+                f'Requested model "{self.model}" is not available in your OpenAI account.'
+            )
+        self._http_client = self._init_http_client
```
matt-bernstein, line 273:
```suggestion
        self._http_client = self._init_http_client()
```
missing parens?

matt-bernstein: also is there a reason this isn't a `@property` like it was in the openai runtimes?

adala/runtimes/_litellm.py
```diff
@@ -0,0 +1,488 @@
...
+                        row,
+                        input_template,
+                        instructions_template,
+                        suffix,
+                        extra_fields,
+                    ),
+                    axis=1,
+                ).tolist()
+
+                responses = await async_concurrent_create_completion(
```
matt-bernstein, line 338: mind adding a todo to replace this with `litellm.acompletion` (or w/e the syntax is) and remove `async_concurrent_create_completion` and `async_create_completion`?

adala/runtimes/_litellm.py
```diff
@@ -0,0 +1,488 @@
...
+    splitter: Optional[str] = None
+    timeout: Optional[int] = 10
+
+    def _init_http_client(self) -> httpx.AsyncClient:
+        client = (
+            httpx.AsyncClient(
+                timeout=self.timeout,
+            ),
+        )
+        litellm.aclient_session = client
```
matt-bernstein, line 261: a little awk that if you create 2 runtimes with different timeouts, the second will overwrite the first's timeout bc this is global, no? it's an ok amount of tech debt but should have a comment about it

matt-bernstein (approved):

adala/runtimes/_openai.py
matt-bernstein: yes

adala/runtimes/_openai.py
matt-bernstein: this is an acceptable level of breakage imo

adala/runtimes/_litellm.py
hakan458: Yeah how did it work at all with missing parens?

pakelley: huh, yeah that is weird, I have definitely been running this locally...

adala/runtimes/_litellm.py
pakelley: wait, what are the two runtimes? You mean the litellm one and the openai subclass?

matt-bernstein: I mean if you create 
```
r1 = LiteLLMAsyncChatRuntime(timeout=10, ...)
r2 = LiteLLMAsyncChatRuntime(timeout=5, ...)
```
then both will be given a timeout of 5 bc the global var is overwritten, right?

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/e7fa146950068ec628e9ad812e11ef8ce38f365b is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/551cd3f2f10a5999dd5afa3b5d05fcd818da8f2f is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/ac2bcdd947afb25d4401af2ec6a4ade40270c2f2 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/3f05f0db3ca046d258e6774ab397b61465db2af8 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/db31b9ca4921ed93d78632371ca4c36ec1053d7e is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/34bc88a7a71ad3006dd0452fdd2d14daf22cc1f0 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/2fe074231a6dacdb2d80327b4111afe4f3427a58 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/db5814ce029fd59c733fe8f85c8c06309e16efc9 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/43257b71c1054681fd83815d1eb4ae7b404cb765 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/b3938ba626960991e15ef9ffb77019b2d3770227 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/9ccf412c569ea02ec1c0525e2fdc47005767ea33 is created.

adala/runtimes/_litellm.py
pakelley: btw, passing the timeout directly to `completion` now, so this won't be an issue anymore :)

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/dd3bb09d473c0ddf8f5edb87ac5830ae402c36d6 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/4d01267ac743b0004e456892f5b1337ff6d58415 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/b074c3df6feff8b754470401966015bebcd3480c is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/acf9614bd8557a6e70267aa352993e694610f75b is created.

PR #162, opened 2024-07-25
feat: RND-105: Support instructor in LLM calls

niklub:
**Updates**:
- refactor LLM calls in a separate file `llm.py` (sync and async)
- create pydantic schema generator based on `output_template` + `field_schema`
- LLM response now contains `data: Dict` field instead of `text`, to provide structured generation (`text` is left for compatibility and will be removed after syncing with server and LSE)
- Some refactoring of tests - moving tests for LLM calls in a separate `test_llm.py` file
- adala error details return full traceback instead of single error string

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,7 +54,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest ]
-        python-version: ['3.9', '3.10', '3.11']
+        # python-version: ['3.9', '3.10', '3.11']
+        python-version: ['3.11']
 
     steps:
       - uses: actions/checkout@v4
--- a/adala/memories/vectordb.py
+++ b/adala/memories/vectordb.py
@@ -1,7 +1,3 @@
-# __import__('pysqlite3')
-# import sys
-# sys.modules['sqlite3'] = sys.modules.pop('pysqlite3')
-
 import chromadb
 import hashlib
 from .base import Memory
--- a/adala/memories/vectordb.py
+++ b/adala/memories/vectordb.py
@@ -36,6 +36,9 @@ class VectorDBMemory(Memory):
         self.remember_many([observation], [data])
 
     def remember_many(self, observations: List[str], data: List[Dict]):
+        # filter None values from each item in `data`
+        data = [{k: v for k, v in d.items() if v is not None} for d in data]
+
         self._collection.add(
             ids=[self.create_unique_id(o) for o in observations],
             documents=observations,
--- a/adala/memories/vectordb.py
+++ b/adala/memories/vectordb.py
@@ -1,3 +1,7 @@
+# __import__('pysqlite3')
+# import sys
+# sys.modules['sqlite3'] = sys.modules.pop('pysqlite3')
+
 import chromadb
 import hashlib
 from .base import Memory
--- a/adala/runtimes/_litellm.py
+++ b/adala/runtimes/_litellm.py
@@ -55,6 +55,16 @@ class LiteLLMChatRuntime(Runtime):
             )
         return self
 
+    def get_llm_response(self, messages: List[Dict[str, str]]) -> Dict[str, Any]:
+        response = get_llm_response(
+            messages=messages,
+            model=self.model,
+            api_key=self.api_key,
+            max_tokens=self.max_tokens,
+            temperature=self.temperature
+        )
+        return response['data']['text']
+
     def record_to_record(
         self,
         record: Dict[str, str],
@@ -94,7 +104,6 @@ class LiteLLMChatRuntime(Runtime):
             instruction_first=instructions_first,
             model=self.model,
             api_key=self.api_key,
-            messages=messages,
             max_tokens=self.max_tokens,
             temperature=self.temperature,
             response_model=response_model
--- a/adala/runtimes/_litellm.py
+++ b/adala/runtimes/_litellm.py
@@ -83,22 +83,15 @@ class LiteLLMChatRuntime(Runtime):
 
         extra_fields = extra_fields or {}
 
-        # options = {}
-        # for field, schema in field_schema.items():
-        #     if schema.get('type') == 'array':
-        #         options[field] = schema.get('items', {}).get('enum', [])
-
-        messages = [
-            {'role': 'system', 'content': instructions_template},
-            {'role': 'user', 'content': input_template.format(**record, **extra_fields)},
-        ]
-
         response_model = parse_template_to_pydantic_class(
             output_template,
             provided_field_schema=field_schema
-        ),
+        )
 
         response = get_llm_response(
+            user_prompt=input_template.format(**record, **extra_fields),
+            system_prompt=instructions_template,
+            instruction_first=instructions_first,
             model=self.model,
             api_key=self.api_key,
             messages=messages,
@@ -107,33 +100,7 @@ class LiteLLMChatRuntime(Runtime):
             response_model=response_model
         )
 
-        # output_fields = parse_template(
-        #     partial_str_format(output_template, **extra_fields),
-        #     include_texts=True,
-        # )
-
-        # outputs = {}
-        # for output_field in output_fields:
-        #     if output_field['type'] == 'text':
-        #         if user_prompt is not None:
-        #             user_prompt += f"\n{output_field['text']}"
-        #         else:
-        #             user_prompt = output_field['text']
-        #     elif output_field['type'] == 'var':
-        #         name = output_field['text']
-        #         messages.append({'role': 'user', 'content': user_prompt})
-        #         completion_text = self.execute(messages)
-        #         if name in options:
-        #             completion_text = match_options(
-        #                 completion_text, options[name]
-        #             )
-        #         outputs[name] = completion_text
-        #         messages.append(
-        #             {'role': 'assistant', 'content': completion_text}
-        #         )
-        #         user_prompt = None
-        #
-        # return outputs
+        return response['data']
 
 
 class AsyncLiteLLMChatRuntime(AsyncRuntime):
@@ -186,20 +153,6 @@ class AsyncLiteLLMChatRuntime(AsyncRuntime):
             )
         return self
 
-    def _prepare_prompt(
-        self,
-        row,
-        input_template: str,
-        instructions_template: str,
-        suffix: str,
-        extra_fields: dict,
-    ) -> Dict[str, str]:
-        """Prepare input prompt for OpenAI API from the row of the dataframe"""
-        return {
-            'system': instructions_template,
-            'user': input_template.format(**row, **extra_fields) + suffix,
-        }
-
     async def batch_to_batch(
         self,
         batch: InternalDataFrame,
@@ -212,79 +165,27 @@ class AsyncLiteLLMChatRuntime(AsyncRuntime):
     ) -> InternalDataFrame:
         """Execute batch of requests with async calls to OpenAI API"""
 
-        extra_fields = extra_fields or {}
-        field_schema = field_schema or {}
-
-        options = {}
-        for field, schema in field_schema.items():
-            if schema.get('type') == 'array':
-                options[field] = schema.get('items', {}).get('enum', [])
-
-        output_fields = parse_template(
-            partial_str_format(output_template, **extra_fields),
-            include_texts=True,
+        response_model = parse_template_to_pydantic_class(
+            output_template,
+            provided_field_schema=field_schema
         )
 
-        if len(output_fields) > 2:
-            raise NotImplementedError('Only one output field is supported')
+        extra_fields = extra_fields or {}
+        user_prompts = batch.apply(lambda row: input_template.format(**row, **extra_fields), axis=1).tolist()
 
-        suffix = ''
-        outputs = []
-        for output_field in output_fields:
-            if output_field['type'] == 'text':
-                suffix += output_field['text']
+        responses = await parallel_async_get_llm_response(
+            user_prompts=user_prompts,
+            system_prompt=instructions_template,
+            instruction_first=instructions_first,
+            max_tokens=self.max_tokens,
+            temperature=self.temperature,
+            model=self.model,
+            api_key=self.api_key,
+            timeout=self.timeout,
+            response_model=response_model
+        )
 
-            elif output_field['type'] == 'var':
-                name = output_field['text']
-                # prepare prompts
-                prompts = batch.apply(
-                    lambda row: self._prepare_prompt(
-                        row,
-                        input_template,
-                        instructions_template,
-                        suffix,
-                        extra_fields,
-                    ),
-                    axis=1,
-                ).tolist()
-
-                # TODO refactor to remove async_concurrent_create_completion and async_create_completion
-                responses = await parallel_async_get_llm_response(
-                    prompts=prompts,
-                    instruction_first=instructions_first,
-                    max_tokens=self.max_tokens,
-                    temperature=self.temperature,
-                    model=self.model,
-                    api_key=self.api_key,
-                    timeout=self.timeout,
-                )
-
-                # parse responses, optionally match it with options
-                for prompt, response in zip(prompts, responses):
-                    completion_text = response.pop('text')
-                    if self.verbose:
-                        if response['error'] is not None:
-                            print_error(
-                                f'Prompt: {prompt}\nLiteLLM API error: {response}'
-                            )
-                        else:
-                            print(
-                                f'Prompt: {prompt}\nLiteLLM API response: {completion_text}'
-                            )
-                    if name in options and completion_text is not None:
-                        completion_text = match_options(
-                            completion_text, options[name]
-                        )
-                    # still technically possible to have a name collision here
-                    # with the error, message, details fields `name in options`
-                    # is only `True` for categorical variables, but is never
-                    # `True` for freeform text generation
-                    response[name] = completion_text
-                    outputs.append(response)
-
-        # TODO: note that this doesn't work for multiple output fields e.g.
-        #       `Output {output1} and Output {output2}`
-        output_df = InternalDataFrame(outputs)
+        output_df = InternalDataFrame(responses)
         return output_df.set_index(batch.index)
 
     async def record_to_record(
--- a/adala/runtimes/_litellm.py
+++ b/adala/runtimes/_litellm.py
@@ -1,4 +1,3 @@
-import asyncio
 import logging
 from typing import Any, Dict, List, Optional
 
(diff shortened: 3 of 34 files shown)


robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/025a1a9ff5641281dd7b302445b8eb6b9bec7fc7 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/2e38785ba3c8a178a387b46e48f94c55b56cc1dd is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/3070c534fde0486c81b571b6840278853a6cfa19 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/4d8edba15cdf82c18889420e1486b7dfdd35e6f0 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/66785f8ea9c2f98a66733a45075f6091204d6bde is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/04dff98fc68f98835e122cef18d9f36575805f3f is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/7fd202ab0695bcf38753b4125d829a324b7097a3 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/7c769a79bb3af7ae1dda731c4ee6939ff29417d0 is created.

.github/workflows/tests.yml
```diff
@@ -54,7 +54,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest ]
-        python-version: ['3.9', '3.10', '3.11']
+        # python-version: ['3.9', '3.10', '3.11']
```
niklub, line 57: Temporary disabling these tests due to incompatibility in vcr stubs, addressing later in https://humansignal.atlassian.net/browse/RND-106

matt-bernstein: can you enable 3.11 so at least some tests can be run?

adala/memories/vectordb.py
```diff
@@ -1,3 +1,7 @@
+# __import__('pysqlite3')
+# import sys
+# sys.modules['sqlite3'] = sys.modules.pop('pysqlite3')
+
```
matt-bernstein, line 4: dead code

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+
+instructor_client = instructor.from_litellm(litellm.completion)
+async_instructor_client = instructor.from_litellm(litellm.acompletion)
+
+
+class LLMResponse(BaseModel):
+    data: Dict = Field(default_factory=dict)
+    adala_error: bool = Field(default=False, serialization_alias='_adala_error')
+    adala_message: str = Field(default=None, serialization_alias='_adala_message')
+    adala_details: str = Field(default=None, serialization_alias='_adala_details')
```
matt-bernstein, line 17: definitely appreciate using a typed object here instead of a dataframe :+1:

adala/runtimes/_litellm.py
```diff
@@ -144,25 +55,15 @@ def init_runtime(self) -> 'Runtime':
...
             max_tokens=self.max_tokens,
-            temperature=self.temperature,
+            temperature=self.temperature
         )
-        completion_text = completion.choices[0].message.content
-
-        if self.verbose:
-            print(f'LiteLLM response: {completion_text}')
-        return completion_text
+        return response['data']['_completion_text']
```
matt-bernstein, line 66: error handling - is the invariant from LSEHandler preserved here that `data == None` iff `adala_error == True`? Can `_completion_text` ever not exist in `data`?

adala/utils/llm.py
matt-bernstein: can fields within `data` also be added to the schema here?

adala/runtimes/_litellm.py
```diff
@@ -191,43 +92,26 @@ def record_to_record(
...
-                    )
-                outputs[name] = completion_text
-                messages.append(
-                    {'role': 'assistant', 'content': completion_text}
-                )
-                user_prompt = None
 
-        return outputs
+        response.update(**response.get('data', {}))
+        response.pop('data', None)
```
matt-bernstein, line 113: I see this is flattening `data`, why?

adala/utils/llm.py
matt-bernstein: different objects for constrained and unconstrained generation might be clearer

adala/runtimes/_litellm.py
```diff
@@ -191,43 +92,26 @@ def record_to_record(
...
-                completion_text = self.execute(messages)
-                if name in options:
-                    completion_text = match_options(
-                        completion_text, options[name]
-                    )
-                outputs[name] = completion_text
-                messages.append(
-                    {'role': 'assistant', 'content': completion_text}
-                )
-                user_prompt = None
```
matt-bernstein, line 228: happy to see this logic go, it was confusing :+1:

adala/skills/collection/classification.py
```diff
@@ -29,15 +29,7 @@ def validate_labels(self):
                 )
 
             self.field_schema[labels_field] = {
-                "type": "array",
-                "items": {"type": "string", "enum": labels},
+                "type": "string",
+                "description": "The classification label.",
+                "enum": labels
```
matt-bernstein, line 34: have you tried running the server tests? wondering if the server payload from LSE needs to be updated to account for this

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def get_messages(user_prompt: str, system_prompt: Optional[str] = None, instruction_first: bool = True):
+    messages = [{'role': 'user', 'content': user_prompt}]
+    if system_prompt:
+        if instruction_first:
+            messages.insert(0, {'role': 'system', 'content': system_prompt})
+        else:
+            messages[0]['content'] += system_prompt
```
matt-bernstein, line 26: would this only be used if the runtime did not support a system prompt, is there another reason why you might do this? should system_prompt be added to the beginning of the message or passed as a separate user message anyway?

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+            response_model=response_model,
+            timeout=timeout,
+        )
+        response.data = instructor_response.model_dump(by_alias=True)
+    except Exception as e:
+        response.adala_error = True
+        response.adala_message = type(e).__name__
+        response.adala_details = traceback.format_exc()
+    finally:
+        return response.model_dump(by_alias=True)
```
matt-bernstein, line 109: see above comment about separate data models for constrained and nonconstrained responses

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
+import asyncio
+import instructor
+import litellm
+import traceback
+import multiprocessing as mp
+from typing import Optional, Dict, Any, List, Type, Union
+from pydantic import BaseModel, Field
+
+instructor_client = instructor.from_litellm(litellm.completion)
+async_instructor_client = instructor.from_litellm(litellm.acompletion)
```
matt-bernstein, line 10: very slick :100:

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+                temperature=temperature,
+                timeout=timeout,
+                instruction_first=instruction_first,
+                response_model=response_model,
+            )
+        )
+        for user_prompt in user_prompts
+    ]
+    responses = await asyncio.gather(*tasks)
+    return responses
```
matt-bernstein, line 140: with httpx, this did not need a separate codepath, you could just pass `concurrent_clients` to the Client from the Runtime like this:
```
            http_client=httpx.AsyncClient(
                limits=httpx.Limits(
                    max_connections=self.concurrent_clients,
                    max_keepalive_connections=self.concurrent_clients,
                ),
                timeout=self.timeout,
            ),
```
is there a way to pass this option through litellm?

adala/utils/pydantic_generator.py
matt-bernstein, line 1: I bet there are 3rd party utils to do all of this, but if this impl works, we don't need to explore them

tests/test_skills.py
```diff
@@ -303,6 +353,10 @@ def test_linear_skillset():
     )
 
     agent.learn(learning_iterations=2)
+    assert agent.skills[
+               "skill_0"].instructions == '"Given a category and its context, provide a list of items that belong to this category. The context will help in generating a more accurate list. Do not provide definitions or explanations, just list the items."'  # noqa: E501
+    # TODO: not learned with 2 iterations, need to increase learning_iterations
+    assert agent.skills["skill_1"].instructions == '...'
```
matt-bernstein, line 359: @pakelley you were running into a bug where `...` was generated as the openai response due to missing a field in the runtime or something, right? has that happened here?

pakelley (commented):
submitting initial comments in case you come ack to look at Matt's comments :)

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+from typing import Optional, Dict, Any, List, Type, Union
+from pydantic import BaseModel, Field
+
+instructor_client = instructor.from_litellm(litellm.completion)
+async_instructor_client = instructor.from_litellm(litellm.acompletion)
+
+
+class LLMResponse(BaseModel):
+    data: Dict = Field(default_factory=dict)
+    adala_error: bool = Field(default=False, serialization_alias='_adala_error')
```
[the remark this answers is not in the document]
matt-bernstein: yes, the original reason for naming `_adala_*` is because they were columns inside a Dataframe, to avoid name collisions. Now they're properly typed so that's not an issue.

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+                api_key=api_key,
+                messages=messages,
+                max_tokens=max_tokens,
+                temperature=temperature,
+                timeout=timeout,
+            )
+            completion_text = completion.choices[0].message.content
+            response.data['_completion_text'] = completion_text
+        except Exception as e:
+            response.adala_error = True
```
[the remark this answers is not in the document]
pakelley: I figure we can use the LLMResponse object for all of our internal logic, and hold off on doing the model_dump until we have to serialize it for something external. Maybe I'm missing something here though.

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+        response.adala_message = type(e).__name__
+        response.adala_details = traceback.format_exc()
+    finally:
+        return response.model_dump(by_alias=True)
+
+
+async def parallel_async_get_llm_response(
+    user_prompts: List[str],
+    system_prompt: Optional[str] = None,
+    model: Optional[str] = 'gpt-4o-mini',
```
pakelley, line 115: nit: shouldn't be optional, it just has a default (same with instruction_first, max_tokens, and temperature)

adala/utils/llm.py
pakelley: sort of, you can set the client at the module level (this is what I was doing at first, but decided to get away from it since at the time it seemed we only needed to pass the timeout through the client).
I'm skeptical of assigning the http client at the module level though, it seems error prone... If it works though, I'm all good with it

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+    ]
+    responses = await asyncio.gather(*tasks)
+    return responses
+
+
+def get_llm_response(
+    user_prompt: Optional[str] = None,
+    system_prompt: Optional[str] = None,
+    messages: Optional[List[Dict[str, str]]] = None,
+    model: Optional[str] = 'gpt-4o-mini',
```
pakelley, line 147: same thing w/ optional params as above

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+        response_model: Pydantic model to constrain the LLM generated response. If not provided, the raw completion text will be returned.  # noqa
+        max_tokens: Maximum tokens to generate.
+        temperature: Temperature for sampling.
+        timeout: Timeout in seconds.
+
+    Returns:
+        Dict[str, Any]: OpenAI response or error message.
+    """
+
+    if not user_prompt and not messages:
```
pakelley, line 63: This api feels clunky, why allow both "user_prompt + system_prompt" and "messages"? Also doesn't seem we use (or test) passing messages directly anyway

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+    max_tokens: Optional[int] = 1000,
+    temperature: Optional[float] = 0.0
+):
+    pool = mp.Pool(mp.cpu_count())
+    responses = pool.starmap(
+        get_llm_response,
+        [
+            (
+                user_prompt,
+                system_prompt,
```
pakelley, line 222: since this is passing args positionally, wouldn't `model` be passed to `messages` in this case?

adala/utils/parse.py
```diff
@@ -85,3 +87,85 @@ def parse_template(string, include_texts=True) -> List[TemplateChunks]:
         )
 
     return chunks
+
+
+def parse_template_to_pydantic_class(
```
pakelley, line 92: this is neat, love that we can do this now

.github/workflows/tests.yml
niklub: I left [3.11 enabled](https://github.com/HumanSignal/Adala/pull/162/files/d6e5e355877f9688c12587abc6b8f59f42ceffb7#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR58)

adala/utils/llm.py
niklub: > can fields within data also be added to the schema here?

no, because they are user-defined, we don't know them in advance therefore can not specify in model fields

niklub: > different objects for constrained and unconstrained generation might be clearer

good idea, I can support that, however, I don't see any reason why we need unconstrained generation at all, and it can be removed in the future versions

adala/runtimes/_litellm.py
niklub: To comply with the current schema where each field in the output template makes out individual dataframe column, like `Output 1: {output1}, output 2: {output2}` -> output1, output2 columns

adala/utils/llm.py
niklub: I imagine the situation where you have to instruct your LLM AFTER providing the long context, to avoid lost-in-the-middle effect. 
However, I think overall `instructions` parameter can be redundant from the UX prospective and we can let user define the structure of the prompt directly in `input_template`

tests/test_skills.py
niklub: here the effect is due to the "live" test calls that don't reach second skill `skill_1` as `skill_0` remains unoptimized during 2 iterations

adala/utils/llm.py
niklub: I had to leave this for compatibility with `agent.learn()` calls, for example: https://github.com/HumanSignal/Adala/blob/master/adala/skills/_base.py#L318

adala/utils/llm.py
niklub: do you mean leaving it as 
```
model: str = 'gpt-4o-mini'
```
?

adala/utils/llm.py
niklub: anyway, changing this behavior is out of the scope as it is how all skills are currently working.

adala/utils/llm.py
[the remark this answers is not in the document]
niklub: the logic with creating dataframe is still in place, nothing changed on this way. That's why `serialization_alias` is provided

adala/runtimes/_litellm.py
```diff
@@ -144,25 +55,15 @@ def init_runtime(self) -> 'Runtime':
             )
         return self
 
-    def execute(self, messages: List):
-        """
-        Execute LiteLLM request given list of messages in OpenAI API format
-        """
-        if self.verbose:
```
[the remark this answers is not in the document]
niklub: resolved back

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/7d59e18c5155c88653f1e78de436b77e508a847d is created.

adala/utils/llm.py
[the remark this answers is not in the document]
niklub: Done with returning models:
```
class LLMResponse(BaseModel):
    """
    Base class for LLM response.
    """


class ConstrainedLLMResponse(LLMResponse):
    """
    LLM response from constrained generation.
    `data` object contains fields required by the response model.
    """
    data: Dict = Field(default_factory=dict)


class UnconstrainedLLMResponse(LLMResponse):
    """
    LLM response from unconstrained generation.
    `text` field contains raw completion text.
    """
    text: str = Field(default=None)


class ErrorLLMResponse(LLMResponse):
    """
    LLM response in case of error.
    """
    adala_error: bool = Field(default=True, serialization_alias='_adala_error')
    adala_message: str = Field(default=None, serialization_alias='_adala_message')
    adala_details: str = Field(default=None, serialization_alias='_adala_details')
```

however, logic with putting everything in dataframe is still there by design.

adala/utils/llm.py
```diff
@@ -0,0 +1,235 @@
...
+            completion = await litellm.acompletion(
+                model=model,
+                api_key=api_key,
+                messages=messages,
+                max_tokens=max_tokens,
+                temperature=temperature,
+                timeout=timeout,
+            )
+            completion_text = completion.choices[0].message.content
+            response.data['_completion_text'] = completion_text
```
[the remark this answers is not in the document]
niklub: resolved with https://github.com/HumanSignal/Adala/pull/162#discussion_r1692436952

adala/utils/pydantic_generator.py
niklub: do you have some tool in mind?

adala/utils/llm.py
niklub: Resolved this with https://github.com/HumanSignal/Adala/pull/162#discussion_r1692436952

adala/runtimes/_litellm.py
niklub: resolved it with https://github.com/HumanSignal/Adala/pull/162#discussion_r1692436952

adala/utils/llm.py
niklub: resolved with https://github.com/HumanSignal/Adala/pull/162#discussion_r1692436952

adala/utils/llm.py
niklub: made a TODO to deprecate this "raw" call entirely, removing it from `learn()`

adala/utils/llm.py
niklub: fixed

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/547f3bb0137c33459b3649214211184a8c98251e is created.

farioas (approved):

adala/utils/llm.py
niklub: fixed

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/b3523140af3a2268e3be440289837b639e4f98cc is created.

adala/utils/pydantic_generator.py
niklub: based on the latest comment from [here](https://github.com/pydantic/pydantic/issues/643#issuecomment-583344600), it is either supported with cli or `create_model` is recommended, same as the current implementation. Other tools based on my search rely on writing a python code string and execute it, which imo is not secure enough. Let's stick with the current implementation and replace it as soon as it will be natively supported by pydantic

matt-bernstein: thanks for looking into it, makes sense. I'd remembered seeing this nifty snippet to generate a pydantic model from a function definition and wondered if it could be extended to json schemas https://x.com/HamelHusain/status/1781836071392952560

adala/skills/collection/classification.py
niklub: Checked with the live server
image

Skipping one test as it fails in the master branch

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/c4fb86779fcdf94988b14a6452bfc0e5164170c5 is created.

adala/utils/llm.py
pakelley: If they are, fantastic, but I've been bitten by that assumption (with libraries that seem focused on being "prod-ready") before

adala/utils/llm.py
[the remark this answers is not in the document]
pakelley: yeah, saw the serialization alias. Figure it'd be nice to (at least internally) have a more helpful name, but I don't mind too much either way

pakelley (approved):

matt-bernstein (approved):

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/50686b5b00d68e264f7ee2d07839a614f46ab1c4 is created.

PR #175, opened 2024-08-07
feat: DIA-1303: [adala BE] Skip RetryErrors when surfacing model provider errors in adala

matt-bernstein:
forward the root cause error after an instructor RetryError


robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/bfeaa86424bca4178b62973cb8aece4341c84cc8 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/71f903bf836bb696220b85918ce1ca82d222ed2d is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/4da86aa6add99bcd8d0d66bc0f51a2f3502a7172 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/6934f375c2079a97c11994f69a1964d4405bcbb5 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/703ec5f7e09031124f4ca4550843f701521a4d45 is created.

adala/runtimes/_litellm.py
```diff
@@ -158,12 +159,31 @@ def record_to_record(
...
             )
+        except InstructorRetryException as e:
+            # get root cause error from retries
+            n_attempts = e.n_attempts
+            e = e.__cause__.last_attempt.exception()
+            logger.error(f"Inference error {e} after {n_attempts=}")
+            error_message = type(e).__name__
+            error_details = str(e)
+            tb = traceback.format_exc()
+            if self.verbose:
```
pakelley, line 170: I know I fought for keeping this in a while ago, but if we're doing a logger.error anyway, I figure this is somewhat redundant at that point? Feels more natural to use logging levels, and just print the traceback at the debug level, I think. If we care about rich (iirc that's all print_error is adding to print), then we can use the rick logging handler.

adala/runtimes/_litellm.py
```diff
@@ -283,12 +303,33 @@ async def batch_to_batch(
...
         for response in responses:
-            if isinstance(response, Exception):
-                error_message = type(response).__name__
-                # error_details = str(response)
-                error_details = traceback.format_exc()
+            if isinstance(response, InstructorRetryException):
+                e = response
+                # get root cause error from retries
+                n_attempts = e.n_attempts
+                e = e.__cause__.last_attempt.exception()
```
pakelley, line 310: On the fence about this, but it is definitely tempting to just make a function to convert to `error_dct` from `e` at least... the logging can still be determined in each section for itself imo, but seems like the sort of conversion we may end up with in a couple places?

pakelley: Like an `instructor_retry_exception_repr` or `format_instructor_retry_exception` kind of thing. Definitely not necessary though, whatever seems right to you

tests/test_llm.py
```diff
@@ -32,10 +32,30 @@ def test_llm_sync():
...
+        record={"input_name": "Carla", "input_age": 25},
+        input_template="My name is {input_name} and I am {input_age} years old.",
+        instructions_template="",
+        output_template="name: {name}, age: {age}",
+    )
+
+    expected_result = {
+        '_adala_error': True,
+        '_adala_message': 'AuthenticationError',
+        '_adala_details': "litellm.AuthenticationError: AuthenticationError: OpenAIException - Error code: 401 - {'error': {'message': 'Incorrect API key provided: fake_api_key. You can find your API key at https://platform.openai.com/account/api-keys.', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_api_key'}}"
```
pakelley, line 49: maybe just worth asserting that something like "AuthenticationError" is in the details string? Would hate for this to start failing for minor format changes from litellm

tests/test_llm.py
```diff
@@ -32,10 +32,30 @@ def test_llm_sync():
...
+    }
+    assert result == expected_result
+
 
 @pytest.mark.vcr
 def test_llm_async():
 
+    # test success
```
pakelley, line 57: stray coment?

pakelley (approved):

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/f21345dc39c5527cad6771ab0ff6344482681069 is created.

adala/runtimes/_litellm.py
matt-bernstein: yep, getting rid of the verbose setting here

maybe later it'd make sense to just have that set the log level...

adala/runtimes/_litellm.py
matt-bernstein: yep, did this. Kept the branch on error type outside the function for clarity

tests/test_llm.py
matt-bernstein: as opposed to testing failure :)

tests/test_llm.py
matt-bernstein: gonna hold off on this for now bc displaying the full error in the test is useful for docs, but def open to not asserting on it

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/77052e87c930ad1518bbe3851350eef5fe91e999 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/216cd0a6a71426e27ae4efe84ecce2ec85736de0 is created.

PR #168, opened 2024-07-29
feat: RND-109: Add EntityExtraction skill in Adala

niklub:
Make response_model parameter top-level for skill definition, to simplify development avoiding redundant params: intructions, output_template, field_schema.

Add EntityExtraction skill to extract the list of entities with specified labels

The skill supports 2 different modes:

**with labels**

`EntityExtraction(input_template='….', labels=['Orgnanization', ‘Product’])`

given the text document and instruction provided with input_template, returns the payload like: 

```
[[{'quote_string': 'Apple Inc.',
'label': 'Organization'},
{'quote_string': 'American multinational technology company',
'label': 'Product'}], ... ]
```

**without labels:**

`EntityExtraction(input_template='…')`

the payload based on the input_template instruction is the following:

```
[[{'quote_string': 'Apple Inc.'},
{'quote_string': 'American multinational technology company'},
{'quote_string': 'consumer electronics'},
{'quote_string': 'computer software'}, ...]
```

PR #176, opened 2024-08-08
fix: DIA-1328: [adala BE] learn method is not working with structured generation

matt-bernstein:
There is now a mismatch between the `predictions` and `feedback` arrays in `Agent.learn()` where the predictions come out as pieces of the pydantic model generated from the skill's schema, and the feedback comes out as plaintext (the previous behavior), which causes false negative matches.

As `learn()` is currently implemented, I believe the previous behavior is correct, since the field names (eg `Label`) interfere with the model's understanding of the problem. However, if `learn()` was using structured generation end2end, for example like this: https://python.useinstructor.com/examples/examples/, then the new behavior could be an improvement. However, the scope of this ticket is just to resolve the mismatch.

See the comments in the latest commit for an example.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -98,7 +98,7 @@ jobs:
             core.setOutput("docker-tags", docker_tags);
 
       - name: Push Docker image
-        uses: docker/build-push-action@v6.5.0
+        uses: docker/build-push-action@v6.6.1
         id: docker_build_and_push
         with:
           context: .
--- a/adala/agents/base.py
+++ b/adala/agents/base.py
@@ -346,7 +346,22 @@ class Agent(BaseModel, ABC):
             )
 
             inputs = self.environment.get_data_batch(batch_size=batch_size)
+
+            # TODO more thought-through fix here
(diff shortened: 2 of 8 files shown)


robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/bd729b219fd57bd12c71473593ac89f63a53059b is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/db85c836ee718cc00d4257f8b2a53898ff19ed63 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/ea7f8d7e626acb0d4a85ae15e014cb1a25ed6916 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/a37d9f07334179c5c2d49b7c39522810ecd8585c is created.

matt-bernstein:
fixed in #168

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/820d95c5d042d5347023fbaa8c0d19b4cdcd794d is created.

PR #214, opened 2024-09-23
feat: DIA-1402: V1-Submit Prompt auto-refinement job

matt-bernstein:
Add a prompt improvement skill and an endpoint to call it through the server

TODO
[x] validation
[x] test coverage
[x] clean up our data models - started in this new codepath, can extend to existing codepaths later


robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/14031fcdd71ee23269a54bfcd78b810c2cbaa9fc is created.

server/prompt_improvement_skill.py
```diff
@@ -0,0 +1,64 @@
...
+    reasoning: str
+    improved_user_prompt: str
+    # NOTE: not exposed in LSE yet, so default is always used. Should improve this as well when we expose it.
+    # improved_system_prompt: str
+
+
+    @field_validator("improved_user_prompt", mode="after")
+    def validate_used_variables(cls, value: str) -> str:
+
+        start_variable_idx = value.find("{")
```
pakelley, line 20: annoyingly, I think we're probably best off using a regex for this kind of thing. We'd have false positives using this for e.g. "blah {blah blah} blah"

pakelley (approved):
overall lgtm

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/b7ba401d08ae771e19944b715eefe0fbfa6b8b12 is created.

server/prompt_improvement_skill.py
niklub: agree, we have a special function for this case: https://github.com/HumanSignal/Adala/blob/master/adala/utils/parse.py#L39

server/prompt_improvement_skill.py
```diff
@@ -0,0 +1,64 @@
...
+
+def get_prompt_improvement_skill(input_variables: List[str]) -> TextGenerationSkill:
+
+    # setting this dynamically - used to validate the improved prompt
+    PromptImprovementSkillResponseModel._input_variables = input_variables
+    
+    prompt_improvement_skill = TextGenerationSkill(
+        name="prompt_improvement",
+        instructions="Improve the user prompt for the provided LLM model to complete the task using the provided input variables, with the provided user prompt as a starting point. Variables can be accessed in the user prompt using the format {variable_name} (only the variable values are used, not their names). Make sure your prompt produces output that will continue to conform to the provided json schema. Provide your reasoning for the changes you made to the prompt.",
+        input_template='''
```
niklub, line 53: Let's move it to a separarate file, so we could keep refining it easily.
A couple of a prompt eng best practices that we could incorporate right away:

- split each variable in a well defined block, for example: `Model: {model}` --> `## Model\n{model}\n\n`
- add fewshot examples
- add `# Instructions:` or `# Follow the steps:` section with a numbered list of instructions
- add `# Things to avoid:` section where we patch the mistakes 

prompt engineering for refinement is out of the scope for sure, but moving `input_template` in a separate file and leaving `instructions` empty would allow us to easily iterate on that in the future updates

server/app.py
```diff
@@ -307,6 +311,81 @@ async def ready(redis_conn: Redis = Depends(_get_redis_conn)):
...
+    Improve a given prompt using the specified model and variables.
+
+    Args:
+        request (ImprovedPromptRequest): The request model for improving a prompt.
+
+    Returns:
+        Response: Response model for prompt improvement skill
+    """
+    
+    inputs = get_prompt_improvement_inputs(request.student_skill, request.input_variables, request.student_model)
```
niklub, line 367: why not making it internal function of `Agent` ?

matt-bernstein: good point. The info that doesn't fit in the Agent is the `input_variables`, because that's a list of all variables (from the project) that _could_ be used in the prompt regardless of if they currently are, and the Environment in the agent would not be used. But it's close enough that it could go there.

server/prompt_improvement_skill.py
matt-bernstein: I want to keep all the deps for the PromptImprovementSkill together, not sure putting just the input template in a separate file would be any cleaner.

But one thing I want to get your opinion on re prompt engineering is the trick of factoring out all the prompt contents that don't contain variables into the system prompt, and leaving the variables in the user prompt. This way we can easily use prompt caching on the system prompt, and it would also change the performance in some way - not sure if positive or negative, but I suspect positive, and I wanted to trial it here first, and if it works here bring it to LSE. wdyt?

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/e1cd62d405c7b2ddd1e4808baa2e1c8a4c50654c is created.

server/prompt_improvement_skill.py
matt-bernstein: done

server/app.py
matt-bernstein: done

server/prompt_improvement_skill.py
matt-bernstein: Updated the prompt. Keeping as separate system and user prompts for now, lmk if you have a strong opinion about that

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/9076e602021526e96fd6c3acc67cde7b8418a3c6 is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/9d931af115a155b97ca0e1c0c93876a08151bb4f is created.

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/6e0bdd789ddc14f21ef158c9f071efab51cd3a40 is created.

adala/skills/collection/prompt_improvement.py
```diff
@@ -0,0 +1,149 @@
...
+        "response_json_schema": student_skill.response_model.model_json_schema(),
+    }
+
+
+def get_prompt_improvement_skill(input_variables: List[str]) -> TextGenerationSkill:
+
+    # setting this dynamically - used to validate the improved prompt
+    PromptImprovementSkillResponseModel._input_variables = input_variables
+
+    prompt_improvement_skill = TextGenerationSkill(
```
niklub, line 87: We have `AnalysisSkill` that effectively transforms table into a single generated string

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/823f5854931f56aac9b0b1217bec0b4864421e64 is created.

tests/test_server.py
```diff
@@ -474,3 +475,59 @@ def test_streaming_azure(client):
             assert (
                 actual_output == expected_output
             ), "adala did not return expected output"
+
+
+# can't use vcr here because server makes async requests
+@pytest.mark.use_openai
```
niklub, line 481: it is not clear why responses can't be mocked here. I'd prefer to avoid manual pytest runs and use it alongside other checks, if it is currently implied by use_openai mark here.

niklub (changes requested):

adala/skills/_base.py
```diff
@@ -548,30 +589,22 @@ class AnalysisSkill(Skill):
...
-
-        Args:
-            input (InternalDataFrame): The input data to be processed.
-            runtime (Runtime): The runtime instance to be used for processing.
+    def _iter_over_chunks(self, input: InternalDataFrame, chunk_size: Optional[int] = None):
 
-        Returns:
-            InternalSeries: The record containing the analysis results.
-        """
+        if input.empty:
```
niklub, line 598: this one is a wip, the chunk iteration flow will be improved within the current PR

server/app.py
```diff
@@ -2,23 +2,28 @@
...
 import os
 import json
+import pandas as pd
 
 import fastapi
 from fastapi import Request, status
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse
 from adala.agents import Agent
+from adala.skills import Skill
```
hakan458, line 12: dont need this anymore I think

hakan458 (approved):

adala/skills/collection/prompt_improvement.py
```diff
@@ -0,0 +1,182 @@
...
+    new_prompt_content: str = Field(..., description="The new content for the prompt")
+
+    # model_config = ConfigDict(
+    #     # omit other fields
+    #     extra="ignore",
+    #     # guard against name collisions with other fields
+    #     populate_by_name=False,
+    # )
+
+    # @field_validator("new_prompt_content", mode="after")
```
niklub, line 27: I'm going to restore this code somehow as we need to post-validate the generating. But the private var `_input_variable` is not working with instructor

tests/test_server.py
niklub: Don't need this test as it almost duplicates the `test_refine_skill`, will put more variations there and remove this one

adala/skills/collection/prompt_improvement.py
niklub: https://humansignal.atlassian.net/browse/DIA-1489

adala/skills/_base.py
niklub: https://humansignal.atlassian.net/browse/DIA-1489

niklub (approved):

PR #220, opened 2024-10-01
fix: DIA-1441: RecursionError better traceback logging

hakan458:
Currently when we are printing the traceback for exceptions caught we were using `traceback.format_exc` which uses `*sys.exc_info()` instead of accepting an exception object itself. For some reason in this flow, this was returning None. Now using `traceback.format_exception()` which DOES take an exception object argument, and will return the traceback as a list of strings. Current just printing it (like `traceback.print_exception()` but might be useful to send back to LSE in the future

### DIFF
--- a/adala/runtimes/_litellm.py
+++ b/adala/runtimes/_litellm.py
@@ -84,7 +84,9 @@ def _format_error_dict(e: Exception) -> dict:
 def _log_llm_exception(e) -> dict:
     dct = _format_error_dict(e)
     base_error = f"Inference error {dct['_adala_message']}"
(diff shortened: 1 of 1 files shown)


robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/40bcd0e891fb165fa0f428bfa035a3b89130fc8d is created.

matt-bernstein (approved):

robot-ci-heartex:
Commit https://github.com/HumanSignal/infra/commit/90fa6536cec34229bf1e919be92c04abebc55c23 is created.

PR #225, opened 2024-10-07
feat: DIA-1384: add cost estimate endpoint

pakelley:


server/app.py
```diff
@@ -210,6 +224,88 @@ async def submit_batch(batch: BatchData):
     return Response[BatchSubmitted](data=BatchSubmitted(job_id=batch.job_id))
 
 
+def get_prompt_tokens(string: str, model: str, output_fields: List[str]) -> int:
+    user_tokens = litellm.token_counter(model=model, text=string)
```
hakan458, line 228: Of course the endpoint needs to be here but what do you think about putting the litellm stuff in litellm runtime? Seems cleaner to have it all in one file, wdyt? Also not using other things that are part of the runtime here, just one simple function, so not a clear choice to me, wondering if you thought about it

server/app.py
```diff
@@ -210,6 +224,88 @@ async def submit_batch(batch: BatchData):
...
+
+
+def get_completion_tokens(model: str, output_fields: List[str]) -> int:
+    max_tokens = litellm.get_model_info(model=model, custom_llm_provider="openai").get(
+        "max_tokens", None
+    )
+    if not max_tokens:
+        raise ValueError
+    # extremely rough heuristic, from testing on some anecdotal examples
+    return min(max_tokens, 4 * len(output_fields))
```
hakan458, line 242: why 4x ?

server/app.py
```diff
@@ -210,6 +224,88 @@ async def submit_batch(batch: BatchData):
...
+
+
+def _estimate_cost(user_prompt: str, model: str, output_fields: List[str]):
+    prompt_tokens = get_prompt_tokens(user_prompt, model, output_fields)
+    completion_tokens = get_completion_tokens(model, output_fields)
+    prompt_cost, completion_cost = litellm.cost_per_token(
+        model="gpt-3.5-turbo",
```
hakan458, line 249: why is model hard coded here?

server/app.py
```diff
@@ -210,6 +224,88 @@ async def submit_batch(batch: BatchData):
...
+
+    return prompt_cost, completion_cost, total_cost
+
+
+@app.post("/estimate-cost", response_model=Response[CostEstimate])
+async def estimate_cost(
+    request: CostEstimateRequest,
+):
+    """
+    Submits a batch of data to an existing streaming job.
```
hakan458, line 263: update this copy pasted docstring :)

server/app.py
pakelley: yeah, this is the "rough heuristic" from that comment... it just seems to be roughly what the tokens come out to.
Not my favorite, but basically a slight improvement over the "static number of tokens for a system prompt" estimate we talked about the other day

server/app.py
pakelley: yeah, I was on the fence about it. It kinda feels like we don't need the separate litellm runtime since it's sort of subsuming everything else, but I agree that if we have it this probably makes sense to put in there. I'll move it and if we decide to break that up later we can decide the best place for it then.

server/app.py
pakelley: missed it, I'll fix that, thanks

hakan458 (approved):

adala/agents/base.py
```diff
@@ -216,6 +216,16 @@ def get_teacher_runtime(self, runtime: Optional[str] = None) -> Runtime:
             )
         return runtime
 
+    def get_skills(self) -> List[Skill]:
```
niklub, line 219: why do we need this function? Skills must be always Dict mapping from skill name to Skill (there is a `skills_validator` function to prevalidate that), so we can always iterate like
```
for skill_name in agent.skills.get_skill_names():
     skill = agent.skills[skill_name]
```

adala/runtimes/_litellm.py
```diff
@@ -527,6 +528,73 @@ async def record_to_record(
...
+                completion_cost_usd=cumulative_completion_cost,
+                total_cost_usd=cumulative_total_cost,
+            )
+
+        except Exception as e:
+            logger.error("Failed to estimate cost: %s", e)
+            return CostEstimate(
+                prompt_cost_usd=None,
+                completion_cost_usd=None,
+                total_cost_usd=None,
```
niklub, line 595: do we need to include error details in response?

adala/runtimes/base.py
```diff
@@ -191,6 +216,11 @@ def record_to_batch(
             response_model=response_model,
         )
 
+    def get_cost_estimate(
```
niklub, line 219: we can use @abc.abstractmethod

adala/runtimes/base.py
```diff
@@ -12,6 +12,31 @@
 tqdm.pandas()
 
 
+class CostEstimate(BaseModel):
```
niklub, line 15: why not using a unique type defined in SDK, here and in LSE?

tests/test_cost_estimation.py
```diff
@@ -0,0 +1,81 @@
...
+    assert isinstance(cost_estimate.total_cost_usd, float)
+    assert np.isclose(
+        cost_estimate.total_cost_usd,
+        cost_estimate.prompt_cost_usd + cost_estimate.completion_cost_usd,
+    )
+
+
+@pytest.mark.use_openai
+def test_estimate_cost_endpoint():
+    test_client = TestClient(app)
```
niklub, line 37: use `client` fixture instead, to be consistent with other tests (e.g. `test_server`)

niklub (changes requested):

adala/runtimes/base.py
pakelley: I didn't want to force every runtime to define this, esp since I'm not sure how do-able it is for images yet

adala/runtimes/base.py
pakelley: We don't want to depend on the sdk anywhere other than the LSE-specific result handler.
We were told that adala is meant to be runnable separately from LSE, but lmk if we're planning to couple them together now and we can start doing things like this.

adala/runtimes/_litellm.py
niklub: at least now we see the error with Azure that we should surface to the user as an actionable message

adala/agents/base.py
niklub: > so we can't count on this always being a SkillSet 
we can because of https://github.com/HumanSignal/Adala/blob/master/adala/agents/base.py#L109

adala/runtimes/base.py
pakelley: Well, it is about LSE if we have to add LSE as a dependency. Are you not proposing that the LSE SDK becomes a dependency of adala? (outside of the LSE handler, which is the only place we currently depend on the SDK)

adala/agents/base.py
pakelley: Can we change the type sig of `skills` in `Agent` then?
And still, like I said in my first comment, it seems simpler to have a `SkillSet.get_skills` than `SkillSet.get_skill_names`, esp since `SkillSet.get_skill_names` doesn't seem to be used anywhere. I'll go ahead and change that and the type sig of `Agent.skills`

adala/runtimes/_litellm.py
pakelley: yeah exactly, we saw that in the Adala logs. Are you saying this is good enough, or are you suggesting you want us to be able to also see that in the LSE logs? Fwiw I don't necessarily think that's more useful than what we have now

pakelley: What "actionable message" would we give the user in this case? How do you think this error should be displayed? I'm assuming we don't want our generic error popup in this case, should the FE determine which errors to surface to the user and which to simply fail and not show an error?

pakelley: I'd propose that the FE just puts a "failed to generate a cost estimate" or blanks out that area rather than popping up an error.

pakelley: We can forward the error message to LSE, if you think it's important to have it in the LSE logs (in addition to the Adala logs)

PR #232, opened 2024-10-18
feat: DIA-1511: Support data batch context for autorefinement

matt-bernstein:
fix /improve-prompt request to not require a jobid, as a job has not yet been created

### DIFF
--- a/adala/agents/base.py
+++ b/adala/agents/base.py
@@ -30,6 +30,7 @@ from adala.utils.logs import (
 )
 from adala.utils.internal_data import InternalDataFrame
 from adala.utils.types import BatchData
+
 logger = logging.getLogger(__name__)
 
 
@@ -400,7 +401,7 @@ class Agent(BaseModel, ABC):
         self,
         skill_name: str,
         input_variables: List[str],
-        batch_data: Optional[BatchData] = None,
+        data: Optional[List[Dict]] = None,
     ) -> ImprovedPromptResponse:
         """
         beta v2 of Agent.learn() that is:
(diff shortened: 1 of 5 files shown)


niklub:
Does this PR introduce any functional changes or is it only about reformatting and renaming?

matt-bernstein:
the only functional change is this
> fix /improve-prompt request to not require a jobid, as a job has not yet been created

which I assume was a bug, unless it's to support some future pathway what I didn't know about

pakelley (approved):

pakelley:
Approved as far as these changes being fine, but to be clear: do we just not actually need any substantive changes (aside from not requiring the job_id) to support this?

matt-bernstein:
> Approved as far as these changes being fine, but to be clear: do we just not actually need any substantive changes (aside from not requiring the job_id) to support this?

That's right. It's a bit hard to follow all the plumbing, but the PromptImprovementSkill is currently implemented as an AnalysisSkill: DataFrame -> Record, where all other inputs go in the static `instructions` field and the GT data batch is the DataFrame. This needs some rethinking in the future if we want the ability to input multiple kinds of example data, such as with and without GT, or positive (correct) vs negative (incorrect) examples. Especially since we don't actually want anything implemented in the AnalysisSkill (just batching, afaik), But that's what we have for now.

matt-bernstein:
> Approved as far as these changes being fine, but to be clear: do we just not actually need any substantive changes (aside from not requiring the job_id) to support this?

That's right. It's a bit hard to follow all the plumbing, but the PromptImprovementSkill is currently implemented as an AnalysisSkill: DataFrame -> Record, where all other inputs go in the static `instructions` field and the GT data batch is the DataFrame. This needs some rethinking in the future if we want the ability to input multiple kinds of example data, such as with and without GT, or positive (correct) vs negative (incorrect) examples. Especially since we don't actually want anything implemented in the AnalysisSkill (just batching, afaik), But that's what we have for now.

niklub (approved):

hakan458 (approved):

PR #240, opened 2024-10-29
fix: DIA-1556: Not getting predictions for all tasks

hakan458:
acks='all' ensures that all replicas have written the message that is produced 

group_id for consumers is to allow multiple consumers to consume from the same topic. In this case we have a 1-to-1 relationship and each group_id should be unique in order to properly track the offset for each topic

manual offset committing in the consumer seems to have solved the issue. 4 in a row runs with 10k/10k outputs. 

We still need the `time.sleep(0.1)` in the submit_batch endpoint, without it we still hit the same offset issue on the consumer side... unclear why exactly.

### DIFF
--- a/adala/environments/kafka.py
+++ b/adala/environments/kafka.py
@@ -55,6 +55,7 @@ class AsyncKafkaEnvironment(AsyncEnvironment):
             self.kafka_input_topic,
             bootstrap_servers=self.kafka_bootstrap_servers,
             value_deserializer=lambda v: json.loads(v.decode("utf-8")),
+            enable_auto_commit=False,
             auto_offset_reset="earliest",
             group_id=self.kafka_input_topic,
         )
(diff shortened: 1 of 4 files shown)


niklub (approved):
Great to see these findings! approved, but please clarify the meaning of each arguments by adding the comments in the code

adala/environments/kafka.py
```diff
@@ -55,14 +55,16 @@ async def initialize(self):
             self.kafka_input_topic,
             bootstrap_servers=self.kafka_bootstrap_servers,
             value_deserializer=lambda v: json.loads(v.decode("utf-8")),
+            enable_auto_commit=False,
```
matt-bernstein, line 58: Should this also be added to the output topic consumer?

matt-bernstein: also, mind leaving a comment about what this + acks=all does?

hakan458: Yeah I will add to output as well. For whatever reason, maybe due to rate of messages sent to output topic, we never hit this issue there. Will leave comments as well yes

PR #243, opened 2024-11-01
fix: DIA-1534: JSON values interpreted as format spec

matt-bernstein:


tests/test_llm.py
```diff
@@ -27,7 +27,7 @@ class Output(BaseModel):
 
     result = runtime.record_to_record(
         record={"input_name": "Carla", "input_age": 25},
-        input_template="My name is {input_name} and I am {input_age:02d} years old with {brackets:.2f} and {brackets2:invalid_format_spec}.",
+        input_template="My name is {input_name} and I am {input_age:02d} years old with {brackets:.2f} and {brackets2:invalid_format_spec} and {input_name:invalid_format_spec}.",
```
niklub, line 30: is it possible to add tests with exactly the prompt that caused the bug?

niklub (approved):

PR #245, opened 2024-11-04
fix: DIA-1584: Use send_and_wait + batches for output topic

hakan458:
Back to using send_and_wait, which while using before we never lost a message in this stage. 

However, we switched to send for speed purposes. To mitigate this, we are now sending batches of records to the output topic, instead of calling send_and_wait once per record. Even if we have the small delay while awaiting the result, that will now happen once per batch (e.g. 50 records) instead of once per record, greatly reducing the time it takes.


server/tasks/stream_inference.py
```diff
@@ -71,7 +71,7 @@ async def run_streaming(
     task_time_limit=settings.task_time_limit_sec,
 )
 def streaming_parent_task(
-    self, agent: Agent, result_handler: ResultHandler, batch_size: int = 10
+    self, agent: Agent, result_handler: ResultHandler, batch_size: int = 1
```
matt-bernstein, line 74: Why set the batch size to 1 here? is it just reverting another recent change in hopes that it's related to the instability?

adala/environments/kafka.py
```diff
@@ -95,12 +95,9 @@ async def message_sender(
...
         record_no = 0
         try:
-            for record in data:
-                await producer.send(topic, value=record)
-                record_no += 1
-                # print_text(f"Sent message: {record} to {topic=}")
+            await producer.send_and_wait(topic, value=data)
             logger.info(
-                f"The number of records sent to topic:{topic}, record_no:{record_no}"
+                f"The number of records sent to topic:{topic}, record_no:{len(data)}"
```
matt-bernstein, line 100: should the input producer in the fastapi endpoint also have this logic to send_and_wait a whole batch at once?

server/tasks/stream_inference.py
pakelley, line 74: Any chance you've tested this in a FB with a reasonably large dataset? My only concern here is that we cause congestion in the DB from doing a separate write for each pred

pakelley, line 74: Any chance you've tested this on an FB with a reasonably large dataset? My only concern is that we create congestion in the DB from writing on each pred

adala/environments/kafka.py
hakan458: for some reason there has never been a single issue observed from the submit_batch endpoint producer -> agent , so I would say no for now

server/tasks/stream_inference.py
hakan458: No, since we are sending "one message" from the producer in the agent now, which actually contains dozens of records usually, I dont want to pull too many at once, and overload LSE endpoints
Batch size of 1 here == tens of records to send to LSE 

before it was 10, which meant we pull 10 individual records. now we pull a batch

hakan458: We are not calling LSE endpoint per record. See comment above

hakan458: To answer your question yes Ive tried with thousands of records on the LSE FB using this adala deployment.

pakelley (approved):

server/tasks/stream_inference.py
matt-bernstein: ok yeah, so the actual batch size is the one set by adala for the consumer+producer in the kafka environment, which iirc defaults to 100

matt-bernstein (approved):

server/tasks/stream_inference.py
hakan458: yeah, it differs each batch though we dont actually wait for the full size

PR #247, opened 2024-11-05
fix: DIA-1523: test coverage for LabelStudioSkill

matt-bernstein:
Fuzz testing using a bunch of different label configs and models and making sure predictions are always valid

Uncovered and fixed 2 bugs:
- NER postprocessing was only done for the first tag in a label config
- text field failed to generate sometimes for NER entities


matt-bernstein:
black made a bunch of formatting changes, ignore the ones in existing tests, there's just one big new test at the bottom

adala/skills/collection/label_studio.py
```diff
@@ -79,10 +80,11 @@ async def aapply(
...
             )
-            ner_tag = self.has_ner_tag()
-            if ner_tag:
+            for ner_tag in self.ner_tags():
                 input_field_name = ner_tag.objects[0].value.lstrip('$')
                 output_field_name = ner_tag.name
                 quote_string_field_name = 'text'
-                output = extract_indices(pd.concat([input, output], axis=1), input_field_name, output_field_name, quote_string_field_name)
+                df = pd.concat([input, output], axis=1)
+                output = validate_output_format_for_ner_tag(df, input_field_name, output_field_name)
```
pakelley, line 88: Could we just add `validate_output_format_for_ner_tag` to the beginning of `extract_indices`? I figure we'd want this validation included in the calls to `extract_indices` in `EntityExtraction.apply` and `EntityExtraction.aaply` too, or is there too much overhead or something?

matt-bernstein: That's a good point for backwards compat, I was thinking that `validate_output_format_for_ner_tag` is explicitly label-studio related and shouldn't be mixed up with generic NER functionality in the same function as extract_indices, but it should be used in the EntityExtraction skill too

pakelley, line 88: nice, now we can take out this additional call to `validate_output_format_for_ner_tag`, but other than that lgtm 👍

pakelley (approved):

adala/skills/collection/label_studio.py
matt-bernstein: Still need it, I added the call to EntityExtraction.extract_indices not the standalone extract_indices

PR #248, opened 2024-11-07
fix: DIA-1584: Allow input topic consumers to wait until the data is ready

niklub:


server/utils.py
```diff
@@ -19,7 +19,7 @@ class Settings(BaseSettings):
 
     kafka_bootstrap_servers: Union[str, List[str]] = "localhost:9093"
     kafka_retention_ms: int = 180000  # 30 minutes
-    kafka_input_consumer_timeout_ms: int = 1500  # 1.5 seconds
+    kafka_input_consumer_timeout_ms: int = 15000  # 15 seconds
```
hakan458, line 22: this will cause the frontend to show "evaluating..." for 15 seconds after we process the last record even, not an option I think 
Also dont think this is an issue, besides in cases where we have more concurrent runs than the number of rqworkers in the LSE env

matt-bernstein:
Closing, problem is fixed elsewhere

PR #253, opened 2024-11-15
fix: DIA-1619: logging and graceful handling for memory leaks

matt-bernstein:
set https://docs.celeryq.dev/en/v5.4.0/userguide/workers.html#max-memory-per-child-setting so that when a child process hits 1GB, which is about 60% of the total memory limit for the whole worker, it is gracefully restarted after the current task finishes. Plus some cleanup and logging.

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -209,7 +209,6 @@ async def submit_batch(batch: BatchData):
             f"Size of batch in bytes received for job_id:{batch.job_id} batch_size:{batch_size}"
         )
     except UnknownTopicOrPartitionError:
-        await producer.stop()
         raise HTTPException(
             status_code=500, detail=f"{topic=} for job {batch.job_id} not found"
         )
--- a/server/tasks/stream_inference.py
+++ b/server/tasks/stream_inference.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import os
+import psutil
 import time
 import traceback
 
@@ -9,6 +10,7 @@ from adala.agents import Agent
 from aiokafka import AIOKafkaConsumer
 from aiokafka.errors import UnknownTopicOrPartitionError
 from celery import Celery
+from celery.signals import worker_process_shutdown, worker_process_init
 from server.handlers.result_handlers import ResultHandler
 from server.utils import (
     Settings,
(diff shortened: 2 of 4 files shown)


server/app.py
```diff
@@ -209,7 +209,6 @@ async def submit_batch(batch: BatchData):
             f"Size of batch in bytes received for job_id:{batch.job_id} batch_size:{batch_size}"
         )
     except UnknownTopicOrPartitionError:
-        await producer.stop()
```
hakan458, line 212: why removed?

server/tasks/stream_inference.py
```diff
@@ -69,6 +94,8 @@ async def run_streaming(
     serializer="pickle",
     on_failure=parent_job_error_handler,
     task_time_limit=settings.task_time_limit_sec,
+    task_ignore_result=True,
```
hakan458, line 97: Hm this doesnt return anything but perhaps not setting this (default is False, so celery does store what the task returns) caused some sort of leak

server/tasks/stream_inference.py
```diff
@@ -21,16 +23,39 @@
...
+settings = Settings()
+
 REDIS_URL = os.getenv("REDIS_URL", "redis://localhost:6379/0")
 app = Celery(
     "worker",
     broker=REDIS_URL,
     backend=REDIS_URL,
     accept_content=["json", "pickle"],
     broker_connection_retry_on_startup=True,
+    worker_max_memory_per_child=settings.celery_worker_max_memory_per_child_kb,
```
hakan458, line 35: How does per-child work? If I run `celery --concurrency 10` will it allow this much memory per child? 
I believe we run with concurrency 1 on infra side and just have more pod replicas instead, but worth checking

server/app.py
matt-bernstein: It's redundant, the `finally` block will execute even if the error is reraised

server/tasks/stream_inference.py
matt-bernstein: Yep that's what I was thinking, but it's stored in redis, so if anything wouldn't the leak be in that container not the worker container? But I set it just in case

server/tasks/stream_inference.py
matt-bernstein: We do run with `--concurrency 10` on infra side as well. Unfortunately there's no analogous setting that tracks memory for the whole worker, not the "child worker"/"process worker", so this will not catch a case where eg there are 4 children each holding 0.5GB which will still OOM. But it's not tenable to set this flag at (1.725GB memory limit / 10 children ~= 172MB), because each child uses ~250MB minimum even on startup. I can see a case for lowering it further than 1GB, but I don't want to have to think about this when changing `--concurrency` too.

hakan458: I just feel that since we are running with 1- "children" this wont actually catch anything :(

matt-bernstein: you mean running with >1 children?

Could set the limit to say 300MB, so as soon as the memory starts growing it gets restarted.  I think it's a good bet that children will get different amounts of usage, so one will hit the limit first, it's not like they'll all leak at exactly the same rate.

hakan458 (approved):

server/tasks/stream_inference.py
hakan458: Yep meant "10" not "1-"

PR #257, opened 2024-11-18
feat: DIA-1573: Track predictions made with prompts

hakan458:
Adding `num_predictions` to the API call (which LSE does not use, its just ignored so no problems) so we can parse this info from BQ (pushed from activity log)


pakelley (approved):

pakelley:
ugh, yeah especially awkward since it's a POST, but works as a workaround for now 😞

hakan458:
> ugh, yeah especially awkward since it's a POST, but works as a workaround for now 😞

Yeah not my favorite. At least it has no effect on the API, its just ignored, but a little sloppy. Deadlines man

PR #259, opened 2024-11-21
fix: DIA-1660: Adala errors are not reported as ConstrainedGenerationError when they should be

matt-bernstein:
Instructor upstream error handling has changed
- ValidationErrors don't contain the string we were checking with multiskill
- instructor bug leads to them sometimes being reported as AttributeError

### DIFF
--- a/adala/runtimes/_litellm.py
+++ b/adala/runtimes/_litellm.py
@@ -306,7 +306,9 @@ class LiteLLMChatRuntime(InstructorClientMixin, Runtime):
             )
 
             # Catch case where the model does not return a properly formatted output
-            if type(e).__name__ == "ValidationError" and "Invalid JSON" in str(e):
+            # AttributeError is an instructor bug: https://github.com/instructor-ai/instructor/pull/1103
+            # > AttributeError: 'NoneType' object has no attribute '_raw_response'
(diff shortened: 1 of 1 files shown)


adala/runtimes/_litellm.py
```diff
@@ -464,7 +466,13 @@ async def batch_to_batch(
...
 
                 # Catch case where the model does not return a properly formatted output
-                if type(e).__name__ == "ValidationError" and "Invalid JSON" in str(e):
+                # AttributeError is an instructor bug: https://github.com/instructor-ai/instructor/pull/1103
+                # > AttributeError: 'NoneType' object has no attribute '_raw_response'
+                if type(e).__name__ in {"ValidationError", "AttributeError"}:
+                    logger.error(
+                        f"Converting error to ConstrainedGenerationError: {str(e)}"
+                    )
+                    logger.error(f"Traceback:\n{traceback.format_exc()}")
```
niklub, line 475: let's keep logger.error for the error string and logger.debug() for the traceback

matt-bernstein: makes sense, done

hakan458 (approved):

PR #264, opened 2024-11-25
feat: DIA-1450: Image Classification support in adala

matt-bernstein:
[x] parse input variables to split message based on media type
[x] plumb up vision runtime
[x] generalize error handling
[x] let LabelStudioSkill set input variable types
[x] add tests

### DIFF
--- a/adala/runtimes/_litellm.py
+++ b/adala/runtimes/_litellm.py
@@ -1,7 +1,9 @@
 import asyncio
 import logging
-from typing import Any, Dict, List, Optional, Type
+from collections import defaultdict
+from typing import Any, Dict, List, Optional, Type, Union
 from functools import cached_property
+from enum import Enum
 import litellm
 from litellm.exceptions import (
     AuthenticationError,
@@ -57,7 +59,8 @@ RETRY_POLICY = dict(
 
 
 def get_messages(
-    user_prompt: str,
+    # user prompt can be a string or a list of multimodal message chunks
+    user_prompt: Union[str, List[Dict[str, str]]],
     system_prompt: Optional[str] = None,
     instruction_first: bool = True,
 ):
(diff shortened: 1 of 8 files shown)


adala/runtimes/_litellm.py
```diff
@@ -57,7 +59,8 @@
 
 
 def get_messages(
-    user_prompt: str,
+    # user prompt can be a string or a list of multimodal message chunks
+    user_prompt: Union[str, List[Dict[str, str]]],
```
niklub, line 63: Can we use `class Message(TypedDict)` instead of `Dict[str, str]` definition?

adala/skills/collection/label_studio.py
```diff
@@ -43,6 +46,15 @@ def ner_tags(self) -> Iterator[ControlTag]:
             if tag.tag.lower() == "labels":
                 yield tag
 
+    def image_tags(self) -> Iterator[ObjectTag]:
+        # check if any image tags are used as input variables
+        object_tag_names = self.allowed_object_tags or list(interface._objects.keys())
+        interface = LabelInterface(self.label_config)
```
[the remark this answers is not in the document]
niklub: no, I meant 
```
@cached_property
def label_interface(self):
     return LabelInterface(self.label_config)
```

adala/runtimes/_litellm.py
```diff
@@ -603,109 +600,194 @@ def get_cost_estimate(
...
+        ]
+    """
+    # Parse template to get field positions and surrounding text
+    parsed = parse_template(input_template)
+    chunks: List[MessageChunk] = []
+
+    current_chunk: Optional[MessageChunk] = None
+
+    def add_to_current_chunk(chunk):
+        nonlocal current_chunk
```
pakelley, line 646: I'm guessing you've tried already, but would definitely like to avoid all this nonlocal business... I'd think we could accomplish the `push_current_chunk` behavior by using a generator function and iterating through that? And then for `add_to_current_chunk` we could accumulate into a var within the generator function, and empty it after `yield`ing?

pakelley (approved):

PR #275, opened 2024-12-09
fix: DIA-1637: Check object tag compatibility for NER tags

hakan458:
We are filtering in LabelStudioSkill.validate_response_model based on `allowed_control_tags` and `allowed_object_tags` but not overwriting `self.label_config` based on this - only `self.field_schema` 

Now checking to make sure any `Labels` tag points to a compatible object tag

### DIFF
--- a/adala/skills/collection/label_studio.py
+++ b/adala/skills/collection/label_studio.py
@@ -54,7 +54,10 @@ class LabelStudioSkill(TransformSkill):
             tag = self.label_interface.get_control(tag_name)
             if tag.tag.lower() in {"labels", "hypertextlabels"}:
                 if self.allowed_object_tags:
-                    if all(object_tag.tag in self.allowed_object_tags for object_tag in tag.objects):
+                    if all(
+                        object_tag.tag in self.allowed_object_tags
+                        for object_tag in tag.objects
+                    ):
                         tags.append(tag)
                 else:
                     tags.append(tag)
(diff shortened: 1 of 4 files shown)


adala/skills/collection/label_studio.py
```diff
@@ -52,7 +52,9 @@ def ner_tags(self) -> List[ControlTag]:
         tags = []
         for tag_name in control_tag_names:
             tag = self.label_interface.get_control(tag_name)
-            if tag.tag.lower() in {"labels", "hypertextlabels"}:
+            if tag.tag.lower() in {"labels", "hypertextlabels"} and any(
```
hakan458, line 55: this should be `all(` not `any(`

adala/skills/collection/label_studio.py
```diff
@@ -53,7 +53,11 @@ def ner_tags(self) -> List[ControlTag]:
         for tag_name in control_tag_names:
             tag = self.label_interface.get_control(tag_name)
             if tag.tag.lower() in {"labels", "hypertextlabels"}:
-                tags.append(tag)
+                if self.allowed_object_tags:
```
pakelley, line 56: I'm trying to remember why we don't just always pass allowed_object_tags... do we have a good reason that this shouldn't always just be set, or have we kinda just stumbled into having to support both of these cases?
(I know we're not gonna fix it as part of this PR if that's the case, just want to make sure I understand here)

pakelley (approved):

adala/skills/collection/label_studio.py
hakan458: https://github.com/HumanSignal/label-studio-enterprise/blob/develop/label_studio_enterprise/lse_ml_models/skills.py#L475

This is the reason I guess. I dont know the actual details of how passing it would drop information, since we only save filtered field_schema and keep the entire label_config anyway

PR #298, opened 2025-01-08
feat: DIA-1715: VertexAI Gemini model support

hakan458:


adala/runtimes/_litellm.py
```diff
@@ -153,20 +153,32 @@ def _get_usage_dict(usage: Usage, model: str) -> Dict:
...
+    """
+    When using litellm.get_model_info() some models are accessed with their provider prefix
+    while others are not.
+
+    This helper function contains logic which resolves this for supported providers
+    """
+    if "/" in model_name:
+        model_name = "/".join(model_name.split("/")[1:])
+    provider = provider.lower()
+    if provider == "vertexai":
```
pakelley, line 166: why do we need to handle this case? Don't we control what's used as the provider? So we could just use "vertex_ai" instead of "vertexai", yeah?

adala/runtimes/_litellm.py
```diff
@@ -153,20 +153,32 @@ def _get_usage_dict(usage: Usage, model: str) -> Dict:
...
 
-class InstructorClientMixin:
+def resolve_litellm_model_and_provider(model_name: str, provider: str):
+    """
+    When using litellm.get_model_info() some models are accessed with their provider prefix
+    while others are not.
+
+    This helper function contains logic which resolves this for supported providers
+    """
+    if "/" in model_name:
```
pakelley, line 163: I figure this should also make sure that the first part is the provider? Otherwise if someone uses this on e.g. model_name="meta/llama3.2", provider="vertex_ai", they'd end up with surprising behavior

adala/runtimes/_litellm.py
```diff
@@ -153,20 +153,32 @@ def _get_usage_dict(usage: Usage, model: str) -> Dict:
     return data
 
 
-class InstructorClientMixin:
+def resolve_litellm_model_and_provider(model_name: str, provider: str):
```
pakelley, line 156: nit: imo it's a bit clearer to name this "normalize_litellm_model_and_provider", wdyt?

adala/runtimes/_litellm.py
```diff
@@ -153,20 +153,32 @@ def _get_usage_dict(usage: Usage, model: str) -> Dict:
...
+    provider = provider.lower()
+    if provider == "vertexai":
+        provider = "vertex_ai"
+
+    return model_name, provider
+
+
+class InstructorClientMixin(BaseModel):
+    
+    instructor_mode: str = "json_mode"
```
pakelley, line 174: could we make this `instructor_mode: instructor.Mode = instructor.Mode.JSON` to give it a better-defined type?

adala/runtimes/_litellm.py
```diff
@@ -261,7 +272,7 @@ def init_runtime(self) -> "Runtime":
                 model=self.model,
                 max_tokens=self.max_tokens,
                 temperature=self.temperature,
-                seed=self.seed,
+                # seed=self.seed,
```
pakelley, line 275: why are we getting rid of the seed?

adala/runtimes/_litellm.py
```diff
@@ -553,9 +563,12 @@ def _get_prompt_tokens(string: str, model: str, output_fields: List[str]) -> int
         return user_tokens + system_tokens
 
     @staticmethod
-    def _get_completion_tokens(model: str, output_fields: Optional[List[str]]) -> int:
+    def _get_completion_tokens(
+        model: str, output_fields: Optional[List[str]], provider: str
+    ) -> int:
+        model, provider = resolve_litellm_model_and_provider(model, provider)
```
pakelley, line 569: why do we have to normalize this at completion time? Can we get these into a canonical format before we store them, so we don't have to deal with this here?

adala/runtimes/_litellm.py
```diff
@@ -729,8 +751,9 @@ class AsyncLiteLLMVisionRuntime(AsyncLiteLLMChatRuntime):
 
     def init_runtime(self) -> "Runtime":
         super().init_runtime()
-        if not litellm.supports_vision(self.model):
-            raise ValueError(f"Model {self.model} does not support vision")
+        # model_name = self.model
```
pakelley, line 754: can we leave a comment explaining why we have to do this? And also, ideally, reminding us that we can bring it back after the next release of litellm? (Since the vertex models have the appropriate supports_vision attr now)

pakelley: Also, maybe we could at least still run this for non-vertex models?

server/app.py
```diff
@@ -216,6 +243,99 @@ async def submit_batch(batch: BatchData):
     return Response[BatchSubmitted](data=BatchSubmitted(job_id=batch.job_id))
 
 
+@app.post("/validate-connection", response_model=Response[ValidateConnectionResponse])
+async def validate_connection(request: ValidateConnectionRequest):
+    multi_model_provider_models = {
```
pakelley, line 248: could we name this something like `multi_model_provider_test_models`? Just so it's clear what they're being used for

adala/runtimes/_litellm.py
matt-bernstein: It breaks gemini models, and we can't yet assume params are set correctly during init. Will add a comment about it

adala/runtimes/_litellm.py
matt-bernstein: It has to be de/serialized as part of the Agent json blob passed from LSE, so not sure if that would work. Can test it

adala/runtimes/_litellm.py
```diff
@@ -153,20 +153,32 @@ def _get_usage_dict(usage: Usage, model: str) -> Dict:
...
-class InstructorClientMixin:
+def resolve_litellm_model_and_provider(model_name: str, provider: str):
+    """
+    When using litellm.get_model_info() some models are accessed with their provider prefix
+    while others are not.
+
+    This helper function contains logic which resolves this for supported providers
+    """
+    if "/" in model_name:
+        model_name = "/".join(model_name.split("/")[1:])
```
niklub, line 164:
```suggestion
        model_name = model_name.split('/', 1)[1]
```

adala/runtimes/_litellm.py
hakan458: ugh. currently we dont support any models like this but yeah... not sure how to really check this deterministically though

adala/runtimes/_litellm.py
hakan458: store them where in LSE? Not going to work

adala/runtimes/_litellm.py
hakan458: yes

adala/runtimes/_litellm.py
matt-bernstein: Doesn't work unfortunately

adala/runtimes/_litellm.py
hakan458: comes from here, feels wrong formatting it differently than the others 
https://github.com/HumanSignal/label-studio/pull/6865/files#diff-ea4ce962841ad8637bb0e2aecc79ea8ba18ae29b512a2f0841014169aac058afR14

Id rather keep this admittedly stupid logic in this one function

adala/runtimes/_litellm.py
hakan458: wont happen actually. this `model_name` comes from LSE where we always prefix it with `vertex_ai/` for example

adala/runtimes/_litellm.py
hakan458: updated

server/app.py
hakan458: updated

adala/runtimes/_litellm.py
hakan458: updated

adala/runtimes/_litellm.py
pakelley: Not the hill I'll die on, but seems like it'd be nice to have our names just match litellm in general since the name is somewhat arbitrary. If it bugs you though, I really don't mind this logic that much.

pakelley (approved):

PR #334, opened 2025-02-12
fix: DIA-1911: Gracefully handle large Kafka messages

hakan458:
Even with increased max message size compared to default, it is possible to hit MessageTooSizeLarge errors. Now depending on the size of the message we will split it into as many chunks as we need and send them one by one 

Removing redis from native docker compose as we have it in LSE and don't need another one here.

### DIFF
--- a/docker-compose.native.yml
+++ b/docker-compose.native.yml
@@ -24,9 +24,5 @@ services:
       - KAFKA_KRAFT_CLUSTER_ID=MkU3OEVBNTcwNTJENDM2Qk
       - KAFKA_CFG_AUTO_CREATE_TOPICS_ENABLE=false
       - KAFKA_MESSAGE_MAX_BYTES=3000000
-  redis:
-    image: redis:alpine
-    ports:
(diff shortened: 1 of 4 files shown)


docker-compose.native.yml
```diff
@@ -24,9 +24,5 @@ services:
       - KAFKA_KRAFT_CLUSTER_ID=MkU3OEVBNTcwNTJENDM2Qk
       - KAFKA_CFG_AUTO_CREATE_TOPICS_ENABLE=false
       - KAFKA_MESSAGE_MAX_BYTES=3000000
-  redis:
-    image: redis:alpine
-    ports:
-      - "6379:6379"
-    healthcheck:
-      test: [ "CMD", "redis-cli", "ping" ]
```
jombooth, line 32: guessing this is an intentional removal? Would recommend calling it out in the description if so

jombooth (approved):

pakelley (approved):

PR #341, opened 2025-02-26
fix: update client libs and cache completion model

matt-bernstein:
- update instructor, openai, our fork of litellm
- add caching of completion model/canonical model name to AsyncLitellmRuntime and use it for usage tracking
  - will also use this for cost estimation, but will do it in Azure PR to avoid merge conflicts because there's a bunch of extra logic added there


matt-bernstein:
lib updates cause tests to fail, there are some breaking changes
completion model caching paused pending refactor

PR #346, opened 2025-02-28
fix: DIA-1986: New model providers don't bubble up task-level errors

matt-bernstein:


adala/runtimes/_litellm.py
```diff
@@ -190,18 +175,18 @@ def _openai_client(self):
     def _check_client(self):
         """Make this synchronous"""
         client = InstructorClientMixin(**self.model_dump()).client
-        run_instructor_with_messages(
-            client=client,
+        response = client.chat.completions.create(
```
hakan458, line 178: maybe add the same comment from sync version here

adala/runtimes/_litellm.py
```diff
@@ -232,6 +217,33 @@ class LiteLLMChatRuntime(InstructorClientMixin, Runtime):
...
+    def canonical_model_provider_string(self):
+        """provider_name/model_name"""
+        try:
+            return match_model_provider_string(self.model)
+        except NoModelsFoundError:
+            logger.info(
+                f"Model {self.model} not found in litellm model map for provider {self.provider}. This is likely a single-model deployment."
+            )
+        except Exception as e:
+            logger.exception(
```
hakan458, line 230: probably logger.error right..?

adala/runtimes/_litellm.py
```diff
@@ -232,6 +217,33 @@ class LiteLLMChatRuntime(InstructorClientMixin, Runtime):
...
+        try:
+            resp = self._check_client()
+            return match_model_provider_string(resp.model)
+        except NoModelsFoundError:
+            logger.warning(
+                f"Model {self.model} not found in litellm model map for provider {self.provider}. This is likely a custom model."
+            )
+            return self.model
+        except Exception as e:
+            logger.exception(
```
hakan458, line 242: logger.error

adala/runtimes/_litellm.py
```diff
@@ -342,6 +354,33 @@ def check_concurrency(cls, value) -> int:
             )
         return value
 
+    @property
```
hakan458, line 357: can we please not duplicate this function in 2 places? move it out and have the 2 classes call the same function or something

adala/runtimes/_litellm.py
matt-bernstein: logger.exception provides the full stack trace, and is otherwise the same. It's helpful for debugging

matt-bernstein: if we get here then this is unexpected and we have another edge case to fix, in contrast to NoModeslFoundError which is expected

adala/runtimes/_litellm.py
matt-bernstein: done

hakan458 (approved):

PR #352, opened 2025-03-05
 fix: DIA-1885: edge cases in string parsing

matt-bernstein:
Added new simpler string formatter with state. Should resolve errors in cost estimation, inference, refinement caused by weird prompt/tasks. See updated tests for new behavior.

### DIFF
--- a/adala/skills/collection/entity_extraction.py
+++ b/adala/skills/collection/entity_extraction.py
@@ -32,13 +32,14 @@ def validate_output_format_for_ner_tag(
             continue
         text = row[input_field_name]
         entities = row[output_field_name]
-        for entity in entities:
-            corrected_text = text[entity["start"] : entity["end"]]
-            if entity.get("text") is None:
-                entity["text"] = corrected_text
-            elif entity["text"] != corrected_text:
-                # this seems to happen rarely if at all in testing, but could lead to invalid predictions
-                logger.warning(f"text and indices disagree for a predicted entity")
+        if entities is not None:
+            for entity in entities:
+                corrected_text = text[entity["start"] : entity["end"]]
+                if entity.get("text") is None:
+                    entity["text"] = corrected_text
+                elif entity["text"] != corrected_text:
+                    # this seems to happen rarely if at all in testing, but could lead to invalid predictions
+                    logger.warning(f"text and indices disagree for a predicted entity")
     return df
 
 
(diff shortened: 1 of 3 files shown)


adala/skills/collection/entity_extraction.py
```diff
@@ -32,13 +32,14 @@ def validate_output_format_for_ner_tag(
...
-                # this seems to happen rarely if at all in testing, but could lead to invalid predictions
-                logger.warning(f"text and indices disagree for a predicted entity")
+        if entities is not None:
+            for entity in entities:
+                corrected_text = text[entity["start"] : entity["end"]]
+                if entity.get("text") is None:
+                    entity["text"] = corrected_text
+                elif entity["text"] != corrected_text:
+                    # this seems to happen rarely if at all in testing, but could lead to invalid predictions
+                    logger.warning(f"text and indices disagree for a predicted entity")
```
pakelley, line 42: should we log this as an error, so we have it surfaced in sentry?

matt-bernstein: Sure, good idea

adala/utils/parse.py
```diff
@@ -21,102 +20,78 @@
...
+    }
+
+    processed = format_string
+    for marker, format_str in markers.items():
+        processed = processed.replace(format_str, marker)
+
+    # Escape remaining brackets and restore format strings
+    processed = processed.replace("{", "{{").replace("}", "}}")
+    for marker, format_str in markers.items():
+        processed = processed.replace(marker, format_str)
```
pakelley, line 92: I hate to say it, but could we test that the performance isn't heavily impacted by this? I'm somewhat worried about all of this string formatting
At the very least, can we at least smoke test that this isn't noticeably longer for a few projects?

pakelley: Ideally, including some pathological cases, like Max's project and/or some enhanced prompts

adala/skills/collection/entity_extraction.py
matt-bernstein: Changed. Not logging the actual text to avoid leaking customer data, but not sure how big of a concern that is, can change later if we need to for debugging

adala/utils/parse.py
matt-bernstein: Good idea, will test on Max's project

matt-bernstein: On sample 20 tasks in Max's project, I get 3:35 on this branch vs 3:45 on master with the same outputs (errors on the same 2 tasks), so within expected margin of error for no difference :+1: 
Good callout though

pakelley (approved):

hakan458 (approved):

PR #369, opened 2025-03-31
feat: DIA-2067: Add message type counters in MessageBuilder

niklub:


adala/utils/token_counter.py
matt-bernstein, line 1: Please no abstract classes here :( keep it to the minimum layers of abstraction necessary, here you can just use the litellm token count functions directly https://docs.litellm.ai/docs/completion/token_usage

niklub: We're trying to build the codebase using SOLID principles to resolve potential dependency inversion problems. Specifically, every time you design a new token system, a method of tokenizing, or rely on different ways of counting tokens, you end up with a tightly coupled dependency on litellm. I envision a more sophisticated token management scenario in the future; therefore, decoupling and abstracting this component is essential, as well as  in general following the standards of code development.

matt-bernstein (approved):
can you remove cursorrules from this PR so we can review it separately?

PR #374, opened 2025-04-15
feat: DIA-1996: remove sync kafka library

matt-bernstein:
Don't want to support 2 different libraries' formats for SSL, so to simplify implementation, use `aiokafka` for everything

### DIFF
--- a/server/utils.py
+++ b/server/utils.py
@@ -10,8 +10,9 @@ from typing import List, Union
 import logging
 import os
 from pathlib import Path
-from kafka.admin import KafkaAdminClient, NewTopic
-from kafka.errors import TopicAlreadyExistsError, UnknownTopicOrPartitionError
+from aiokafka.admin import AIOKafkaAdminClient, NewTopic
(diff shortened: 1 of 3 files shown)


server/utils.py
```diff
@@ -55,40 +56,54 @@ def ensure_topic(topic_name: str):
...
+                name=topic_name,
+                num_partitions=1,
+                replication_factor=1,
+                topic_configs={"retention.ms": str(retention_ms)},
+            )
+
+            try:
+                await admin_client.create_topics([topic])
+            except TopicAlreadyExistsError:
+                # we shouldn't hit this case when KAFKA_CFG_AUTO_CREATE_TOPICS=false unless there is a legitimate name collision, so should raise here after testing
```
pakelley, line 77: did you test this and verify whether we can raise here?

pakelley (approved):

server/utils.py
matt-bernstein: I think we should be able to, but deployed envs don't even use this setting, so not worth changing now imo

PR #390, opened 2025-05-01
fix: DIA-1996: make settings variables less permissive

matt-bernstein:
Previously, due to using subsettings, environment variables without the appropriate prefix would shadow ones with the prefix. For example, if `REDIS_URL` is not set, `URL -> REDIS_URL`.

This change prevents picking up unintended env vars.

### DIFF
--- a/server/utils.py
+++ b/server/utils.py
@@ -1,4 +1,5 @@
 from urllib.parse import quote, urlparse, urlunparse, parse_qsl, urlencode
+from pydantic import BaseModel, ConfigDict
 from pydantic_settings import BaseSettings, SettingsConfigDict
 from typing import List, Union, Optional, Literal, Dict, Any
 import logging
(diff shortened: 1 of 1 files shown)


server/utils.py
```diff
@@ -27,7 +28,7 @@ class RedisSettings(BaseSettings):
     ssl_certfile: Optional[str] = None
     ssl_keyfile: Optional[str] = None
 
-    model_config = SettingsConfigDict(
+    model_config = ConfigDict(
```
hakan458, line 31: can you explain the changes?

matt-bernstein: Yep, the subclasses inheriting from BaseSettings meant they would try to look for env vars to pick up on their own independently, now they don't and only Settings checks for env vars

hakan458 (approved):

PR #424, opened 2025-07-07
feat: UTC-159: Cleanup memory leaks in worker pool

hakan458:
- We have been seeing OOMs in adala-worker (celery) in prod after rearhitecture
  - Since all workers are now running long-running tasks, memory is not being fully cleaned up like it was with previous architecture, where the task process would die and OS would reclaim memory 
- There are memory leaks in many places like litellm, instructor, etc and it was not possible to clean up 100% so memory will always grow for each worker after each batch processed 
- Implemented a restart strategy internal to the workers
  - We have a configurable threshold e.g. 350MB per worker 
  - Once this threshold is reached after processing a batch fully, the worker will exit, and celery will automatically spin up a new worker 
- Additional fixes
  - Handling rate limit errors 
  - Fixed issues with many kafka consumers joining group at once causing errors

### DIFF
--- a/server/handlers/result_handlers.py
+++ b/server/handlers/result_handlers.py
@@ -1,7 +1,5 @@
 from typing import Optional, List, Dict
 import json
-import os  # Added for memory tracking
-import psutil  # Added for memory tracking
 from abc import abstractmethod
 from pydantic import BaseModel, Field, ConfigDict, model_validator
 import csv
@@ -12,17 +10,6 @@ from server.utils import init_logger
 
 logger = init_logger(__name__)
 
-# MEMORY TRACKING UTILITY for LSEHandler
-def _log_lse_memory_usage(stage: str) -> float:
-    """Log current memory usage for LSEHandler debugging"""
-    try:
-        process = psutil.Process(os.getpid())
-        memory_mb = process.memory_info().rss / 1024 / 1024
-        logger.info(f"LSEHandler: Memory at {stage}: {memory_mb:.1f}MB")
-        return memory_mb
-    except Exception as e:
-        logger.debug(f"LSEHandler: Error getting memory info: {e}")
-        return 0
 
 try:
     from label_studio_sdk import LabelStudio as LSEClient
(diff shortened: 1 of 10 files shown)


server/handlers/result_handlers.py
```diff
@@ -157,13 +161,13 @@ class LSEHandler(ResultHandler):
...
 
-    model_config = ConfigDict(arbitrary_types_allowed=True)  # for @computed_field
+    model_config = ConfigDict(arbitrary_types_allowed=True)  # for cached_property
 
     api_key: str
     url: str
     modelrun_id: int
 
-    @computed_field
+    @cached_property
```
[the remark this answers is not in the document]
hakan458: We create result handlers through `LSEClientCache` , which is based on unique api_key + model_run_id 
Within each of those in the cache, there is only one client, and that can be cached, so this is ok

server/handlers/result_handlers.py
```diff
@@ -175,9 +179,27 @@ def client(self) -> LSEClient:
...
-        assert self.client.versions.get()
-
-        return self
+        # First attempt without retry to quickly catch auth/config issues
+        try:
+            self.client.versions.get()
+            logger.info(f"LSE client connection verified")
+            return self
+        except Exception as e:
+            # Check if this is a rate limit that should be retried
```
niklub, line 188: lets leaver TODO: to remove this part as we make more robust inter-service communication

server/worker_pool/worker_processor.py
```diff
@@ -160,6 +195,41 @@ async def _check_for_work(self):
...
             logger.error(f"Worker {self.worker_id}: Error checking for work: {e}")
 
+    async def _check_memory_threshold(self):
+        """Check memory usage and trigger restart if needed"""
+
+        try:
+            settings = Settings()
+
+            # Get current memory without logging (to avoid spam)
+            try:
```
niklub, line 204: I think this can be combined into a single utility function to follow DRY

hakan458: done

niklub (approved):

PR #834, opened 2026-07-27
fix: honor required fields in JSON schemas

### LINKED ISSUE #833: JSON Schema converter treats optional properties as required
### Problem

json_schema_to_model() ignores the schema required array and passes Field(...) for every property. As a result, properties that JSON Schema allows callers to omit are rejected by the generated Pydantic model.

### Reproduction

Using a schema with required_name in required and optional_note omitted from required:

    Model = json_schema_to_model({
        "type": "object",
        "properties": {
            "required_name": {"type": "string"},
            "optional_note": {"type": "string"},
        },
        "required": ["required_name"],
    })
    Model(required_name="ok")

Current result:

    ValidationError: optional_note - Field required

Both generated fields report is_required() == True.

### Expected behavior

Only fields named in the JSON Schema required array should be mandatory. Optional fields should remain omittable, while explicit null should still be rejected unless the property schema permits null. The field-schema helper used for skill outputs can continue marking all of its supplied fields as required for backward compatibility.

yaodong-shen:
## Summary

- honor each object schema's required array when creating Pydantic fields
- use declared defaults (or an omittable None default) for non-required properties without making explicit null valid
- preserve the existing contract that all skill response fields are required
- cover top-level and nested required semantics with regression tests

Fixes #833.

## Validation

- black --check adala/utils/pydantic_generator.py tests/test_pydantic_generator.py
- python -m py_compile adala/utils/pydantic_generator.py tests/test_pydantic_generator.py
- pytest -q tests/test_pydantic_generator.py tests/test_skills.py tests/test_label_studio_skill.py tests/test_analysis_skill.py (22 passed, 3 skipped)
- full pytest: 78 passed, 4 skipped, 9 deselected; one unrelated Gemini image VCR test fails because the response lacks content after its image fetch path. The same failure reproduces on clean origin/master.

### DIFF
--- a/adala/utils/pydantic_generator.py
+++ b/adala/utils/pydantic_generator.py
@@ -50,9 +50,12 @@ def json_schema_to_model(json_schema: Dict[str, Any]) -> Type[BaseModel]:
     # `description` is the model class docstring
     model_description = json_schema.get("description", "")
 
+    required_fields = set(json_schema.get("required", []))
     fields_def = {}
     for name, prop in json_schema.get("properties", {}).items():
-        fields_def[name] = json_schema_to_pydantic_field(prop)
+        fields_def[name] = json_schema_to_pydantic_field(
+            prop, required=name in required_fields
+        )
 
     # Create the BaseModel class using create_model().
     model = create_model(model_name, **fields_def)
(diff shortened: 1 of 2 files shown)


yaodong-shen:
The Check, pytests, Docs, and CI/CD Pipeline workflows are waiting for maintainer approval and created no jobs for this external-fork PR. The focused and full local test suites pass apart from the documented unrelated Gemini baseline failure. Could a maintainer approve the workflow runs when convenient?

